PR #21, opened 2020-12-11
Updated ocdb with new use of database components, reorg of .proj files.

d-haldane:


utils/generic-components.stanza
```diff
@@ -105,12 +121,51 @@ public defn closest-std-val (v:Double, tol:Double) :
...
+; ========================================================
+; Generic ceramic capacitor from database (of 600k options)
+public defn ceramic-cap (params:Tuple<KeyValue>) :
+  val default-params = ["type" => "ceramic" 
+                        "mounting" => "smd" 
+                        "case" => get-valid-pkg-list()
+                        "minimum_quantity" => 1
+                        "min-rated-voltage" => MIN-CAP-VOLTAGE
+                        "min-stock" => 5 * DESIGN-QUANTITY]
+  to-esir(Capacitor(to-tuple $ cat-all([default-params params])))
```
PhilippeFerreiraDeSousa, line 136: Usually `cat(list1, list2)` for only 2 lists

utils/generic-components.stanza
```diff
@@ -105,12 +121,51 @@ public defn closest-std-val (v:Double, tol:Double) :
...
+; ========== Generic passive components ==================
+; ========================================================
+; Generic ceramic capacitor from database (of 600k options)
+public defn ceramic-cap (params:Tuple<KeyValue>) :
+  val default-params = ["type" => "ceramic" 
+                        "mounting" => "smd" 
+                        "case" => get-valid-pkg-list()
+                        "minimum_quantity" => 1
+                        "min-rated-voltage" => MIN-CAP-VOLTAGE
+                        "min-stock" => 5 * DESIGN-QUANTITY]
```
PhilippeFerreiraDeSousa, line 135: If you think that you are going to apply min-stock, min-quantity, mounting and case to most request, it maybe worth adding it to default in Capacitor(), Inductor(), Resistor(), or create an overloading function that has those defaults.

But sometimes, too many defaults kills readability, so looks great as it is also

utils/generic-components.stanza
```diff
@@ -120,30 +175,24 @@ public pcb-component gen-cap-cmp (cap:Double, tol:Double, max-v:Double, pkg:Stri
     capacitor => true
     type => "ceramic"
     capacitance => cap
-    tolerance => (tol * 0.01)
```
PhilippeFerreiraDeSousa, line 123: I looked up db-parts:
- tolerance is in % in the emodels as a scalar (symmetric interval)
```
defn emodel-pourcentage-tolerance (value: MinMaxRange|False) -> Double|UNKNOWN :  ; FIXME: Tolerance minmax range not handled properly but emodel is going depricated
  match(value: MinMaxRange) :
    100. * max(value)
  else:
    UNKNOWN
```
- tolerance in properties is only defined for Inductor right now as a tuple unitless:
```
match(tolerance: MinMaxRange): tolerance => [min(tolerance), max(tolerance)]
```
So there is a decision to make on keeping the tuple or a scalar to keep a symmetric interval.

It is a tuple because propoerties only accept primitive types (no MinMaxRange) but there will be a new esir anyway.

utils/db-parts.stanza
```diff
@@ -333,11 +333,11 @@ defmethod to-esir (capacitor: Capacitor) -> VarRef:
       match(esr: ESR) :
         val esr-value = value(esr)
         spice :
-          "[C] {p[1]} tmp {capacitance}"
+          "[C] {p[1]} tmp {capacitance(capacitor)}"
           "[R] tmp {p[2]} {esr-value}"
       else :
         spice :
-          "[C] {p[1]} {p[2]} {capacitance}"
+          "[C] {p[1]} {p[2]} {capacitance(capacitor)}"
```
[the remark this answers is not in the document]
d-haldane: Seems fine...

utils/generic-components.stanza
d-haldane: I'd like to get the parts solver considering the total number of instances of a part when making selections, and not just assume 5. Kinda fine for now because anything you realistically need 100 of per design is most likely in stock in large quantities anyway.

utils/generic-components.stanza
d-haldane: gen-cap-cmp is deprecated for the time being, so I'm not worried about compatibility.

PR #25, opened 2020-12-15
Jb/complete altium power ports

jackbackrack:
Add complete set of Altium power ports with standardized names.


CuppoJava (approved):
Looks good to me.

PR #33, opened 2021-03-05
Adding microchip/ATmega32U4 to components

ariel-hi:
microchip/ATmega32U4
pcb-* definition names need review.
The file contains other definitions -- let's add separate component definitions + files for these?


d-haldane:
The fie needs to follow ocdb standards for component organization - 1 component model per file, optionally with a module. The ATMEGA file needs to be factored out to remove the other components like the diode, usb connector, etc.. and relocate them to other files

PR #34, opened 2021-03-11
Allow the voltage divider solver to accept additional parameters

m-hilgendorf:
This adds an argument, `params` to `make-voltage-divider` that allows a caller to add additional requirements to the solution. For example, if the solution resistors needs to have a higher power rating than default.

### DIFF
--- a/modules/voltage-divider.stanza
+++ b/modules/voltage-divider.stanza
@@ -26,7 +26,9 @@ public defstruct VoltageDividerSolution :
 
 public defn voltage-divider (v-in: [Double, Double, Double],
                              v-out: [Double,Double,Double],
-                             current: Double) -> VoltageDividerSolution|False:
+                             current: Double
+                             params:Tuple<KeyValue<String, ?>>,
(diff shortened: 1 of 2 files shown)


modules/voltage-divider.stanza
```diff
@@ -134,13 +138,18 @@ public defn check-output-voltage-range (v-in: [Double, Double, Double],
...
+  val defaults = [
+    "resistance" => resistance,
+    "tolerance" => tol,
+    "minimum_quantity" => 1,
+    "mounting" => PREFERRED-MOUNTING,
+    "case" => get-valid-pkg-list(),
+    "min-stock" => 5 * DESIGN-QUANTITY
+  ]
+  val params* = to-tuple $ 
+    cat(defaults, filter({not contains?(defaults, _)}, params))
```
PhilippeFerreiraDeSousa, line 151: I tested this, contains? is comparing KeyValue objects instead of keys. Plus I think you want params to override defaults not the opposite. You can use `to-tuple $ to-hashtable<String, ?> $ cat(defaults, params)`

PR #38, opened 2021-04-22
remove board and schematic export for class-a

ariel-hi:


ariel-hi:
`class-a.stanza` has a component without a landpattern and symbol (just spice model). Exporting board and schematic would fail.

ariel-hi:
Two options I saw:
1.  delete the file
2.  remove the board and schematic export, which does render the exported kicad useless. But the file is preserved for when spice models change.

d-haldane (approved):

PR #39, opened 2021-04-23
Merge new additions lift on OCDB master back into dev.

CuppoJava:
New development should be done on `dev` only from now on. This merge grabs some of the new changes on master and puts it back into dev.


d-haldane (approved):
These are fine. APPROVE

PR #41, opened 2021-04-27
Cherry MX model to landpattern

ariel-hi:


d-haldane (commented):
I notice that it's using a KiCAD directory for the 3d-model. Is that intended?

ariel-hi:
I can change the model to match the new export conventions and look like this:
`model3d = model3D("cherry-sw.wrl",
  Vec3D(0., 0., 0.),
  Vec3D(0.3937, 0.3937, 0.3937))
  Vec3D(0.0, 0.0, 0.0))`

If a user includes Cherry MX in their design, after they run their design, they should add `cherry-sw.wrl` to the JITX project's 3d-models folder.

d-haldane:
Sounds like a plan. Same for the Kailh socket model?

ariel-hi:
Done for both models. Also fixed a typo (swapped parenthesis and comma)

d-haldane:
Why is the socket commented out? This would be a good example to show multiple models

ariel-hi:
I don't have the original models, so I wasn't sure if the two models are different ways to represent Cherry MX or if they should be used together to support the landpattern.

ariel-hi:
OK, I looked up these models. Looks like they work in tandem. Uncommenting kailh

d-haldane (approved):

PR #42, opened 2021-04-30
Generated macro validity testing on pull requests

PhilippeFerreiraDeSousa:
# Generated macro validity testing on pull requests

Search for public `pcb-*` macros with 0 or 1 argument of primitive type and make a deftest for each with `print-def` to evaluate them.
Tests for `pcb-module` are not run in this v1 because they can require access to the online part database and take longer to run.

If no annotation is present, default values used are:
```
TYPE_SAMPLE = {
    "Char": '"c"',
    "String": '"hello world"',
    "Byte": '1Y',
    "Int": '1',
    "Long": '1L',
    "Float": '1.0F',
    "Double": '1.0',
    "True|False": 'true'
}
```

The syntax to specify the value to test a macro on is exemplified by:
```
;<test>
n-pins: value1 value2 value3
<test>
public pcb-component component (n-pins:Int) :
```
In this example 3 tests will be generated:
```
deftest test-ocdb_amphenol_minitek127_component:
  print-def(ocdb/amphenol/minitek127/component(value1))

deftest test-ocdb_amphenol_minitek127_component:
  print-def(ocdb/amphenol/minitek127/component(value2))

deftest test-ocdb_amphenol_minitek127_component:
  print-def(ocdb/amphenol/minitek127/component(value3))
```

To use a random even integer between 2 and 100 use:
```
;<test>
n: even positive
<test>
public pcb-landpattern soic127p-landpattern (n:Int) :
```
 
To skip a test do:
```
;<test>skip
n-pins: value1 value2 value3
<test>
public pcb-component component (n-pins:Int) :
```
or
```
;<test>skip<test>
public pcb-component component :
```


.github/jitx-client/scripts/evaluate_pcb_objects.py
```diff
@@ -0,0 +1,235 @@
...
+def or_regex(strings: Iterable[str]) -> str:
+    return "(?:" + "|".join([s.replace("|", "\|") for s in strings]) + ")"
+
+PACKAGE_REGEX = "[\w!?/-]+"
+VARIABLE_REGEX = "[\w!?-]+"
+VALUE_REGEX = '[-+.!?"\w\d]+'
+TYPE_SAMPLE = {
+    "Char": '"c"',
+    "String": '"hello world"',
+    "Byte": 'to-byte(1)',
```
ariel-hi, line 67: `"Byte": '1Y'` can be used here.

components/tag-connect/TC2050-IDC-NL.stanza
```diff
@@ -28,10 +28,10 @@ public pcb-component component :
   description = "Insertion point for TC-2050 connector"
   ; for i in 1 to 11 do :
   ;   pin p[i]
-  port p: pin[1 through 11]
+  port p: pin[1 through 10]
   val sym = header-symbol(10,2)
-  symbol = sym(for i in 1 to 11 do : p[i] => sym.p[i])
-  landpattern = footprint(for i in 1 to 11 do : p[i] => footprint.p[i])
+  symbol = sym(for i in 1 through 10 do : p[i] => sym.p[i])
+  landpattern = footprint(for i in 1 through 10 do : p[i] => footprint.p[i])
```
ariel-hi, line 34: Nice demonstration of `to` vs `through` conventions for 0 vs 1-indexing

utils/land-patterns.stanza
```diff
@@ -7,8 +7,12 @@ defpackage ocdb/land-patterns :
   import jitx/fonts
   import jitx/commands
 
-; ====== Reference Label Helpers ===============================================
+; ====== Argument constraint Helpers ===========================================
+defn ensure-even-positive! (value: Int, name: String):
+  if value < 2 or value % 2 == 1:
+    fatal("%_ only supports a positive even number of pins." % [name])
```
ariel-hi, line 13: What's your methodology for deciding which generators require this check? Based on tests errors?

d-haldane (dismissed):

ariel-hi (commented):
I love this work. Solid testing strategy.

.github/jitx-client/install.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -e
+
+chmod +x /root/.jitx/current/jitx # artifact caching loses permissions somewhere
```
m-hilgendorf, line 4: an alternative (hack) is to `tar` or `zip` the artifact separately, which handles the whole bundle.

PhilippeFerreiraDeSousa: Do you plan on publishing another image with jitx completely installed so that I don't need this install.sh?

m-hilgendorf: The most recent push of the stable client should avoid this

utils/land-patterns.stanza
PhilippeFerreiraDeSousa: I googled sop65 and soic127p to check. But yes they crash for odd integers

.github/jitx-client/scripts/evaluate_pcb_objects.py
PhilippeFerreiraDeSousa: Thx

.github/jitx-client/install.sh
PhilippeFerreiraDeSousa: dev, latest or both tags of jitx-client?

m-hilgendorf: dev

PhilippeFerreiraDeSousa: Great, the CI always pulls the last image so I just removed my jitx installation steps and it stills runs great, thank you!

ariel-hi (approved):

PR #43, opened 2021-05-04
Class-a, ocdb.stanza, gen-ocdb-app, some proj files

ariel-hi:
`class-a.stanza` : Dummy landpattern + symbol to allow this file to be exported.
`ocdb.stanza` : Cleaned up, includes more files. Removed just a few.
`backplanes.stanza` : Removed. Let's define a location for old files that we're interested in keeping.
`gen-ocdb-app.stanza` : Updated package imports. I haven't ran the generator.


scripts/gen-ocdb-app.stanza
```diff
@@ -6,14 +6,13 @@ defpackage fig-app :
...
-  import esir/repl-lib
-  import ocdb/tests/default-harness
+  import jitx
+  import jitx/gen
+  import jitx/commands
+  import ocdb/defaults
   import ocdb/land-patterns
   import ocdb/generic-components
-  import jitpcb/visualizer
+  import jitx/visualizer
```
PhilippeFerreiraDeSousa, line 15: Have you run this? Do you get shiny svgs?

ariel-hi: That's the next step. Just modified the imports for now.

PhilippeFerreiraDeSousa (approved):

d-haldane (commented):
Instead of the dummy landpattern, just use the generator from pin-header, in utils/generic-components

ariel-hi:
@d-haldane I updated power-connector to use pin-header-pkg and box-symbol generators.

d-haldane (approved):

PR #44, opened 2021-05-05
JITX-278: Enabled rounded rectangles

bpellenard:


utils/land-patterns.stanza
```diff
@@ -151,6 +151,34 @@ public defn rect-pth-pad (drill-r:Double, pad-w:Double, pad-h:Double) :
...
   rect-pth-pad(C, drill-w, drill-h, pad-w, pad-h)
 
+pcb-pad roundrect-pth-pad (anchor:Anchor, drill-w:Double, drill-h:Double, pad-w:Double, pad-h:Double, pad-r:Double) :
+  name = to-string("%_ RoundedRectangle PTH Pad (drill = (%_, %_), pad = (%_, %_, %_))" % [
+           anchor, drill-w, drill-h, pad-w, pad-h, pad-r])
+  type = TH
+  if drill-w != drill-h :
+    fatal("roundrect-pth-pad(): RoundedRectangle holes are not supported")
+  
+  shape = RoundedRectangle(pad-w, pad-h, pad-r)
```
bpellenard, line 161: @CuppoJava we do not check validity of pad-w, pad-h and pad-r here, right? We let the RoundedRectangle 'constructor' handle it, right?

utils/land-patterns.stanza
```diff
@@ -151,6 +151,34 @@ public defn rect-pth-pad (drill-r:Double, pad-w:Double, pad-h:Double) :
...
+  layer(Cutout()) = Circle(anchor, drill-w)
+  layer(SolderMask(Top)) = RoundedRectangle(pad-w, pad-h, pad-r)
```
bpellenard, line 163: I followed what was done for ovals. Do we want the same layers to be filled?

utils/land-patterns.stanza
```diff
@@ -151,6 +151,34 @@ public defn rect-pth-pad (drill-r:Double, pad-w:Double, pad-h:Double) :
...
+  roundrect-pth-pad(C, drill-r, drill-r, pad-w, pad-h, pad-r)
+
+public defn roundrect-pth-pad (drill-w:Double, drill-h:Double, pad-w:Double, pad-h:Double, pad-r:Double) :
+  roundrect-pth-pad(C, drill-w, drill-h, pad-w, pad-h, pad-r)
+
+pcb-pad roundrect-smd-pad (anchor:Anchor, w:Double, h:Double, r:Double) :
+  name = to-string("(%_, %_, %_) %_ RoundedRectangle SMD Pad" % [w,h,r,anchor])
+  type = SMD
+  shape = RoundedRectangle(anchor, w, h, r)
+  layer(Paste(Top)) = RoundedRectangle(anchor, w, h, r)
```
bpellenard, line 176: likewise

utils/land-patterns.stanza
m-hilgendorf: I think this is OK for now but we will likely change most of this when improving landpattern generation.

utils/land-patterns.stanza
ariel-hi: If `drill-w != drill-h` then the cutout layer can be a capsule. `rect-pth-pad` and `oval-pth-pad` demonstrate that.

CuppoJava: Yes. It's fine to let the constructor of RoundedRectangle check any preconditions.

bpellenard: Do we want it to be a capsule? I mean, the user asks for a rounded rectangle and we provide only a capsule.

CuppoJava: Is the question: how do we import a Kicad project that contains rounded rectangle drill holes, when we do not support it?

It should either be an error. (This KiCad project cannot be imported for the following reasons.) Or the Kicad importer actually performs a sensible replacement (but this must be made absolutely precise and documented.)

bpellenard: yes exactly. I would vote for an error (blame the caller) when asking for a drill with width different from height. I acknowledge it is currently done as a sensible replacement.

utils/land-patterns.stanza
```diff
@@ -151,6 +151,34 @@ public defn rect-pth-pad (drill-r:Double, pad-w:Double, pad-h:Double) :
 public defn rect-pth-pad (drill-w:Double, drill-h:Double, pad-w:Double, pad-h:Double) :
   rect-pth-pad(C, drill-w, drill-h, pad-w, pad-h)
 
+pcb-pad roundrect-pth-pad (anchor:Anchor, drill-w:Double, drill-h:Double, pad-w:Double, pad-h:Double, pad-r:Double) :
+  name = to-string("%_ RoundedRectangle PTH Pad (drill = (%_, %_), pad = (%_, %_, %_))" % [
+           anchor, drill-w, drill-h, pad-w, pad-h, pad-r])
+  type = TH
+  if drill-w != drill-h :
```
m-hilgendorf, line 158: if these must be equal, we should have one `drill-r` argument instead of two, I think

utils/land-patterns.stanza
CuppoJava: Hmmm, okay I just read through the context some more:

This function happens to be used by the importer, but it is also a standalone OCDB utility that declares a certain pad type. I believe it is common to have rounded-rectangular pads with a capsule-shaped drill hole in the middle. (Please verify with Steve/Duncan, but I believe rounded-rectangular holes are very very rare.) 

So my vote is this pad declaration actually creates a rounded-rectangular pad, but with a capsule-shaped drill hole.

bpellenard: 👍

electricfungi: Rounded rectangles are primarily used for surface mount pads. If a through hole pad uses a rounded rectangle, it will likely take a round hole, or a slotted hole. Of course, there's always corner cases for weird connectors

utils/land-patterns.stanza
ariel-hi: Comment relocated: If `drill-w != drill-h` then the cutout layer can be a capsule. `rect-pth-pad` and `oval-pth-pad` demonstrate that.

utils/land-patterns.stanza
d-haldane: 'Slotted hole' is probably better user level syntax than capsule hole. Should rounded-rectangle just subsume capsules? Capsule seems to just be a special case of a rounded rectangle where r = width/2

bpellenard:
I updated the PR. It is ready for review

m-hilgendorf (approved):
lgtm

PR #45, opened 2021-05-11
Parametric Landpattern generator

m-hilgendorf:


utils/land-patterns/matrix.stanza
```diff
@@ -0,0 +1,102 @@
...
+  column-spacing: Double, ; spacing between columns
+  make-pad: (Int, Int) -> False|JITXDef, ; function to generate the pad based on row/column indices
+  pad-name: (Int, Int) -> Ref,
+):
+  inside pcb-landpattern:
+    val x-offset = 0.5 * to-double(num-columns - 1) * column-spacing
+    val y-offset = 0.5 * to-double(num-rows - 1) * row-spacing
+    for row in 0 to num-rows do:
+      for column in 0 to num-columns do:
+        val pad? = make-pad(row, column)
```
bpellenard, line 94: why `pad?`? I thought ? was for booleans.

utils/land-patterns/pose-arrangement.stanza
```diff
@@ -0,0 +1,127 @@
...
+ 
+defn rotation (degrees:Double) -> Affine:
+  val theta = PI * degrees / 180.0
+  Affine([
+    [cos(theta), -1.0 * sin(theta), 0.0],
+    [sin(theta), cos(theta), 0.0],
+    [0.0, 0.0, 1.0]
+  ])
+
+defn times (p:Point, a:Affine) -> Point:
```
bpellenard, line 72: it is in reverse order: `times (a:Affine, p:Point) -> Point :`

utils/land-patterns/pose-arrangement.stanza
```diff
@@ -0,0 +1,127 @@
...
+  p2[1] = 0.0
+  p2[2] = 0.0
+
+  val A = matrix(a)
+  for i in 0 to 3 do:
+    for j in 0 to 3 do:
+      p2[i] = p2[i] + p1[j] * A[i][j]
+  Point(p2[0], p2[1])
+
+defn times (p:Pose, a:Affine):
```
bpellenard, line 89: likewise for order of arguments

utils/land-patterns/matrix.stanza
```diff
@@ -0,0 +1,102 @@
+; Helpers for creating matrices of pads.
```
bpellenard, line 1: pad-matrix.stanza would be more meaningful

utils/land-patterns/pose-arrangement.stanza
```diff
@@ -0,0 +1,127 @@
+defpackage ocdb/land-patterns/pose-arrangement:
+  import core
+  import collections
+  import math
+  import jitx
+  
+defstruct Affine:
```
bpellenard, line 7: I would move `defstruct Affine` to jitpcb repository such that we can reuse it if needed

utils/land-patterns/pose-arrangement.stanza
```diff
@@ -0,0 +1,127 @@
...
+public defstruct PoseArrangement: 
+  locations: Seqable<Pose>,
+
+defn times (a:Affine, p:PoseArrangement) -> PoseArrangement:
+  times(p, a)
+
+defn times (p:PoseArrangement, a:Affine) -> PoseArrangement:
+  PoseArrangement(seq({_ * a}, locations(p)))
+
+public defn translated (p:PoseArrangement, x:Double, y:Double) -> PoseArrangement:
```
bpellenard, line 102: Maybe that's too much but why not `public defn plus (p:PoseArrangement, xy:[Double, Double]) -> PoseArrangement :`?

or `xy:Array<Double>(2)` if it is supported

utils/land-patterns/pose-arrangement.stanza
```diff
@@ -0,0 +1,127 @@
...
+  PoseArrangement(seq({_ * translate(x, y)}, locations(p)))
+
+public defn scaled (p:PoseArrangement, x:Double, y:Double) -> PoseArrangement:
```
bpellenard, line 105: likewise: `public defn times (p:PoseArrangement, xy:[Double, Double] -> PoseArrangement`

utils/land-patterns/pose-arrangement.stanza
```diff
@@ -0,0 +1,127 @@
...
+
+  var p2 = Array<Double>(2)
+  p2[0] = 0.0
+  p2[1] = 0.0
+  p2[2] = 0.0
+
+  val A = matrix(a)
+  for i in 0 to 3 do:
+    for j in 0 to 3 do:
+      p2[i] = p2[i] + p1[j] * A[i][j]
```
bpellenard, line 86: -> `A[i][j] * p1[j]`
Reading the index is easier

utils/land-patterns/pose-arrangement.stanza
```diff
@@ -0,0 +1,127 @@
...
+defn rotation (degrees:Double) -> Affine:
+  val theta = PI * degrees / 180.0
+  Affine([
+    [cos(theta), -1.0 * sin(theta), 0.0],
+    [sin(theta), cos(theta), 0.0],
+    [0.0, 0.0, 1.0]
+  ])
+
+defn times (p:Point, a:Affine) -> Point:
+  val p1 = Array<Double>(3)
```
bpellenard, line 73: is there a way to do something like `val p1 = Array<Double>(3) [x(p) y(p) 1.0]`
if not that would be great to have it @CuppoJava

utils/land-patterns/matrix.stanza
m-hilgendorf: `pad?` is for things that might be `false`, in this case `False|Pad`

utils/land-patterns/pose-arrangement.stanza
m-hilgendorf: I would like for everything that can be expressed at the library level to be done so. We can always move things around later, but this is for operations on `Pose`s to compose with other operations and should be exposed as much as possible.

utils/land-patterns/pose-arrangement.stanza
m-hilgendorf: The root reason is because in frameworks where I have used these operations, they use the explicit name of the transform (`scale`, `shear`, `translate`, `rotate`, etc). I also find the explicitness more readable than operator overloading.

The difference between the implementation of `times` in here is that under the hood, it is just matrix multiplication. However that is only exposed internally, the functions are not public.

utils/design-vars.stanza
```diff
@@ -13,3 +18,59 @@ public var PREFERRED-MOUNTING = "smd" ; values in ["smd", "through-hole"]
...
+  LevelA
+  LevelB
+  LevelC
+
+; The construction type of the design (rigid, flex)
+public defenum DesignConstructionType:
+  Rigid
+  Flex
+
+; The application domain of the design (consumer, automative, aerospace)
```
electricfungi, line 38: I'm not sure why we need this information for land patterns. I think we should put considerable thought into how to categorize things and what these flags do before adding them. These categories may be too broad to be useful.

utils/land-patterns/pose-arrangement.stanza
bpellenard: why not having both?

```
public defn plus (p:PoseArrangement, xy:[Double, Double]) -> PoseArrangement :
   translated(p, x, y)
```

utils/design-vars.stanza
```diff
@@ -13,3 +18,59 @@ public var PREFERRED-MOUNTING = "smd" ; values in ["smd", "through-hole"]
...
+public defenum DesignApplicationType:
+  Consumer,
+  Automotive,
+;   Aerospace
+
+; The Design Producability corresponds to the IPC 6011 spec for design producability.
```
electricfungi, line 44: IPC-7351

utils/design-vars.stanza
```diff
@@ -13,3 +18,59 @@ public var PREFERRED-MOUNTING = "smd" ; values in ["smd", "through-hole"]
...
+; standards for the designs.
+public var DESIGN-PRODUCIBILITY: DesignProducability = LevelA
+
+; The design construction type refers to the kind of board you are trying to design. 
+; The options are Rigid() or Flex().
+public var DESIGN-CONSTRUCTION-TYPE: DesignConstructionType = Rigid
+
+; The design application type refers to the domain in which your device will function. 
+; The options are False (none), Consumer(), Automotive(), and Aerospace()
+public var DESIGN-APPLICATION-TYPE: False|DesignApplicationType = false
```
electricfungi, line 55: I would defer this kind of flag for now until we have a clear path for how it's used

utils/design-vars.stanza
```diff
@@ -13,3 +18,59 @@ public var PREFERRED-MOUNTING = "smd" ; values in ["smd", "through-hole"]
...
+public var SOLDER-MASK-EXPANSION: Double = 0.005
+
+; Several design rules are required, but it is not yet possible to query them at the 
+; library level.
+public var MIN-SILKSCREEN-WIDTH: Double = 0.1524 ;TODO pull from pcb-rules
+public var MIN-SILKSCREEN-TEXT-HEIGHT: Double = 0.762 ; TODO add to pcb-rules
+public var SILK-SOLDER-MASK-CLEARANCE: Double = 0.15 ; TODO pull from pcb-rules
+public var MIN-TRACE-OUTER: Double = 0.127 
+public var MIN-SPACING-OUTER: Double = 0.127
+
```
electricfungi, line 68: We also may need `public var MIN-SOLDER-MASK-BRIDGE: Double = 0.15` This specifies the minimum width allowed for solder mask (also known as min solder mask sliver or min solder mask dam)

PR #47, opened 2021-05-25
Duncan's too-big demo merge

d-haldane:
This is a bad pull request with a lot lumped in.


m-hilgendorf (changes requested):
`.DS_Store` and `components/analog-devices/.DS_Store` should be deleted.

PR #52, opened 2021-06-04
Parametric Pad Generation

m-hilgendorf:
This is the first PR to add parametric landpattern support to OCDB. 

- Introduces generic pad definitions
- Introduces the "design context" in design-vars
- Refactor pin column/row/grid construction
- Generic landpattern construction from sequences of poses and pads.
- Add or update several pad naming convention helpers.


bpellenard (approved):

designs/tutorial.stanza
```diff
@@ -6,10 +6,19 @@ defpackage ocdb/designs/tutorial :
   import jitx/commands
   import ocdb/defaults
   import ocdb/generic-components
-  
+  import ocdb/design-vars
+
+MIN-PKG = "0603"
+pcb-landpattern
```
d-haldane, line 12: tutorial.stanza should not be modified like this. It's the first design new users see so don't add cruft

utils/land-patterns.stanza
```diff
@@ -301,6 +435,7 @@ public pcb-landpattern dummy-landpattern (n:Int, courtyard:[Double,Double]):
 ; X - pad height
 ; w, h - width and height of courtyard in mm
 public defn make-two-pin-landpattern (Z:Double, G:Double, X:Double, w:Double, h:Double, courtyard?:True|False, polarized?:True|False) :
+  println("making two pin landpattern")
```
d-haldane, line 438: Don't leave print statements in production code.

designs/tutorial.stanza
m-hilgendorf: reverted

utils/land-patterns.stanza
m-hilgendorf: removed

tests/test-connects.stanza
```diff
@@ -0,0 +1,75 @@
...
+  execute-with-error-handler(
+    body,
+    fn (): caught? = true
+  )
+  #EXPECT(caught?)
+
+deftest(ocdb, connects) test-connect-i2c-isolation:
+  pcb-module M1:
+    port p: i2c[[1, 2]]
+    #EXPECT(connect-i2c-isolation(p[1], p[2]) is-not False)
```
bpellenard, line 21: 👍

tests/test-pin-properties-index-field.stanza
```diff
@@ -0,0 +1,22 @@
...
+pcb-component mycomponent :
+  pin-properties :
+    [pin:Ref | pad:Int]
+    [a.x | 0]
+    [a.y | 1]
+    [b.y | 0]
+    [b.x | 1]
+
+deftest test-pin-properties-index-field :
+  #ASSERT(property(mycomponent.a.x.pin-properties-row-index) == 0)
```
bpellenard, line 19: `#EXPECT`

m-hilgendorf: fixed

bpellenard (commented):
It looks good to me from the syntax point of view

bpellenard:
When do you think you will be able to merge this PR?

utils/connects.stanza
```diff
@@ -12,41 +12,86 @@ defpackage ocdb/connects :
...
-          net (y.tx, xcvr.uart.rx)
-          net (y.rx, xcvr.uart.tx)
-          net (y.dtr, xcvr.uart.dtr)
-          net (y.rts, xcvr.uart.rts)
-          schematic-group(xcvr) = xcvr
+  within ensure-bundles!([x, y], [usb-2, uart-with([`dtr, `rts])],  "call connect-ft32"):
+    inside pcb-module :
+      public inst xcvr : ocdb/future-designs/FT232RL/module
+      net (x, xcvr.usb-2)
+      net (y, xcvr.uart)
```
d-haldane, line 95: This is not correct - subtle point, but notice that tx and rx are swapped for the uart connection

utils/debug-utils.stanza
```diff
@@ -478,8 +478,7 @@ public defn db-ftdi-four-comms (pwr:JITXObject, cmp:JITXObject) :
 ; ref : reference to pin or net that test point is attached to
 ; cmp : reference to pcb-component for test point
 ; persist? : test points migrate to bottom layer for production
-public defn add-testpoint (ref:JITXObject, cmp:Instantiable, persist?:True|False, 
-                           top?:True|False) :
+public defn add-testpoint (ref:JITXObject, cmp:Instantiable, persist?:True|False, top?:True|False):
```
d-haldane, line 481: @m-hilgendorf Why was the ability to place testpoints on the bottom removed?

d-haldane (approved):

components/texas-instruments/ISO1540.stanza
```diff
@@ -18,18 +18,16 @@ public pcb-component component :
...
-    [GND2    | 5            | Down    ]
-    [SCL1    | 3            | Left    ]
-    [SCL2    | 6            | Right   ]
-    [SDA1    | 2            | Left    ]
-    [SDA2    | 7            | Right   ]
-    [VCC1    | 1            | Up      ]
-    [VCC2    | 8            | Up      ]
+    [GND1     | 4            | Down    ]
+    [GND2     | 5            | Down    ]
+    [p[1].scl | 3            | Left    ]
```
CuppoJava, line 26: Please confirm this style with Duncan. The code works, but Duncan may want to have the `pcb-component` definition mirror the datasheet as much as possible. You can then move the use of the proper bundle into the `pcb-module` definition.

utils/box-symbol.stanza
```diff
@@ -244,6 +244,22 @@ public defn make-box-symbol (banks:Seqable<BoxBank>) :
 ;=======================================================<doc>
 
 public defn make-box-symbol () :
+  ;Sort the given pins by their .pin-properties-row-index field.
```
CuppoJava, line 247: All the logic here in `box-symbol.stanza` can be removed. It has been superceded by another implementation in a different file.

utils/connects.stanza
```diff
@@ -12,41 +12,90 @@ defpackage ocdb/connects :
...
+      println(IndentedStream(buf), 
+        "Expected %_ for port %_. Found %_." % [
+          name(b), ref(p), 
+          name(b? as Bundle) when b? is-not False else port-type(p)
+        ]
+      )
+    fatal(to-string(buf))
 
+; Connect two i2c lines with a buffer. 
+public defn connect-i2c-isolation (
```
CuppoJava, line 51: Please use idiomatic indenting to be consistent with the rest of the code base:
```
 public defn connect-i2c-isolation (
   x:JITXObject,
   y:JITXObject,
   buffer:Instantiable
 ):
   within ensure-bundles!([x, y], [i2c, i2c], "call connect-isolation"):
     inside pcb-module:
       public inst buf: buffer
       net (x, buf.p[1])
       net (y, buf.p[2])
       buf
```
should be this:
```
 public defn connect-i2c-isolation (x:JITXObject,
                                    y:JITXObject,
                                    buffer:Instantiable):
   within ensure-bundles!([x, y], [i2c, i2c], "call connect-isolation"):
     inside pcb-module:
       public inst buf: buffer
       net (x, buf.p[1])
       net (y, buf.p[2])
       buf
```

utils/connects.stanza
```diff
@@ -12,41 +12,90 @@ defpackage ocdb/connects :
...
-            inst buf : ocdb/texas-instruments/ISO1540/component
-            net (a.sda, buf.SDA1)
-            net (a.scl, buf.SCL1)
-            net (b.sda, buf.SDA2)
-            net (b.scl, buf.SCL2)
-            property(buf.VCC1.requires-power) = true
-            property(buf.VCC1.power-request) = [3.3 0.3 0.05]
+; Helper function to check if the ports have bundles and their assigned
+; bundle matches what we expect
+defn ensure-bundles!<?T> (
```
CuppoJava, line 19: This function is too scary for JITX users. OCDB needs to be written with the perspective that JITX users will scan through it and write more code in the same style. Change the design of this function to be more straightforward.

The current design allows you to write this (at the cost of a very complicated helper function):

```
within ensure-bundles!([x, y], [i2c, i2c], "call connect-isolation"):
   inside pcb-module:
     public inst buf: buffer
     net (x, buf.p[1])
     net (y, buf.p[2])
     buf
```       

Change the design to allow for this style instead.

```
inside pcb-module :
  ensure-bundle-port!(x, i2c)
  ensure-bundle-port!(y, i2c)
  public inst buf: buffer
  net (x, buf.p[1])
  net (y, buf.p[2])
  buf
```

utils/connects.stanza
```diff
@@ -12,41 +12,90 @@ defpackage ocdb/connects :
...
+      buf
+
+; Connect to i2c lines with a buffer, using a default component (TI ISO1540)
+public defn connect-i2c-isolation (
+  x:JITXObject,
+  y:JITXObject,
+):
+  inside pcb-module:
+    val buf = connect-i2c-isolation(x, y, ocdb/texas-instruments/ISO1540/component)
+    match(buf:Instance):
```
CuppoJava, line 70: This `match` is not necessary, because your `connect-i2c-isolation` function is guaranteed to return an `Instance`.

utils/debug-utils.stanza
```diff
@@ -490,17 +491,11 @@ public defn add-testpoint (ref:JITXObject, cmp:Instantiable, persist?:True|False
...
-          schematic-group(tp) = debug-persist-sch
-          layout-group(tp) = debug-persist-layout
-          net (ref, tp.p)
-        else :
-          inst tp : btm-smd-testpoint-cmp()
-          schematic-group(tp) = debug-persist-sch
-          layout-group(tp) = debug-persist-layout
-          net (ref, tp.p)
-
+        inst tp : smd-testpoint-cmp()
```
CuppoJava: Thanks for removing the `top-smd-test-point-cmp` and `bottom-smd-test-point-cmp` definitions. Whether a component is on the top or bottom of a board should not be baked into the component definitions themselves. 

Here is the command that you can use to re-enable support for test points on the bottom.

```
inst tp : smd-testpoint-cmp()
place(tp) on Bottom
```

utils/design-vars.stanza
```diff
@@ -11,5 +11,12 @@ public var MIN-PKG = "0201"
 public var DESIGN-QUANTITY = 100
 public var PREFERRED-MOUNTING = "smd" ; values in ["smd", "through-hole"]
 public var MIN-CAP-VOLTAGE = 10.0 ; Minimum voltage to allow for capacitors
+
 ; For consolidation
 ;public var PREFERRED-MANUFACTURERS = ["Yageo", "Panasonic Electronic Components", "Vishay Dale", "TE Connectivity Passive Product"]
+public var SOLDER-MASK-REGISTRATION: Double = 0.15
```
CuppoJava, line 17: We'll clean this up for you as soon as we get properties in the `pcb-rules` definition. Thanks for staging it strategically.

utils/land-patterns.stanza
```diff
@@ -5,281 +5,406 @@ defpackage ocdb/land-patterns :
...
+    fatal("%_ is not a valid shape for pad geometry: %_." % [name, s])
+  s as Shape
 
 ; ====== Reference Label Helpers ===============================================
-public defn ref-label (x:Double, y:Double, a:Anchor) :
-  inside pcb-landpattern :
-    layer(Silkscreen("values", Top)) = Text(">REF", 0.7056, a, loc(x,y))
-    
+; Returns a reference label text
+public defn ref-label-text (pose:Pose, anchor:Anchor) -> Text:
```
CuppoJava, line 29: To avoid introducing too many concepts to the user, please un-refactor (defactor?) this back into the `ref-label` helper. Here is the reasoning:

If the only function we provide is `ref-label`, then we can provide a clear explanation of what it does, and there isn't anything for the user to complain about.

If we provide two functions: `ref-label` and `ref-label-text`, the user will ask the very reasonable questions:
1) What is the difference between the two?
2) When should I use one versus the other?
3) Do you recommend I follow this style (and factoring) for all of my own utilities as well?

And that is a lot of extra complexity without much gain.

utils/land-patterns.stanza
```diff
@@ -5,281 +5,406 @@ defpackage ocdb/land-patterns :
...
+  if width >= height:
+    ChamferedRectangle(width, height, r)
+  else:
+    ChamferedRectangle(height, width, r)
+
+;==============================================================================
+;=========================== Geometric Helpers ================================
+;==============================================================================
+; Create a rectangle out of lines
+public defn LineRectangle (w:Double, h:Double, xc:Double, yc:Double, line-w:Double) :
```
CuppoJava, line 80: Please do a grep through the most up-to-date version of the codebase. I believe Bertrand might have already relocated these `LineRectangle` helpers to a much more sensible location. If so, then you can remove these definitions.

utils/land-patterns.stanza
```diff
@@ -5,281 +5,406 @@ defpackage ocdb/land-patterns :
...
+    Line(MIN-SILKSCREEN-WIDTH, [center + Point(-0.5 * l, 0.0), center + Point(0.5 * l, 0.0)])
+    Line(MIN-SILKSCREEN-WIDTH, [center + Point(0.0, -0.5 * l), center + Point(0.0, 0.5 * l)])
+  ]
+
+; Create a 1mm x 1mm plus symbol at Point(x, y)
+public defn PlusSymbol (x:Double, y:Double) -> Shape:
+  PlusSymbol(Point(x, y), 1.0)
+
+; Create a row of poses
+public defn pose-row (origin:Pose, count:Int, pitch:Double) -> Seqable<Pose>:
```
CuppoJava, line 118: Please use the existing `row-locs` and `col-locs` utilities instead of redefining `pose-row` and `pose-column`.

utils/land-patterns.stanza
```diff
@@ -5,281 +5,406 @@ defpackage ocdb/land-patterns :
...
+  for n in 0 to count seq:
+    Pose(center(origin) + Point(pitch, 0.0), angle(origin), flip-x?(origin))
+
+; Create a column of poses
+public defn pose-column (origin:Pose, count:Int, pitch:Double) -> Seqable<Pose>:
+  for n in 0 to count seq:
+    Pose(center(origin) + Point(0.0, 0.0), angle(origin), flip-x?(origin))
+
+; Create a matrix of poses by row.
+public defn pose-grid (origin:Pose, num-rows:Int, num-columns:Int, column-pitch:Double, row-pitch:Double,) -> Seqable<Seqable<Pose>>:
```
CuppoJava, line 128: For `pose-grid`, can you ascertain whether the existing `grid-locs` utility can be used? If not, then we need to decide upon the most useful API for this utility and make it standard.

utils/land-patterns.stanza
```diff
@@ -5,281 +5,406 @@ defpackage ocdb/land-patterns :
...
+  for r in 0 to num-rows seq:
+    val row-origin = Pose(center(origin) + Point(0.0, row-pitch), angle(origin), flip-x?(origin))
+    pose-row(row-origin, num-columns, column-pitch)
+
+;==============================================================================
+;================================= Pads =======================================
+;==============================================================================
+; A generic pad, defined by the shape of its pad, solder mask, and hole.
+public pcb-pad generic-pad (
```
CuppoJava, line 137: Remove this `generic-pad` definition. It deviates too far from the style that we want JITX users to write. The original `smd-pad`, `bga-pad`, etc. functions are clearer. If need be, use helpers that populate fields instead of factoring everything into one pad definition.

Here is an example:
```
pcb-pad pth-pad (d:Double) :
  name = "PTH Pad"
  type = TH
  make-hole-and-set-mask(3.0, true)  ;Helper function to create a 3.0 radius hole and populate solder mask fields.
```

utils/land-patterns.stanza
```diff
@@ -5,281 +5,406 @@ defpackage ocdb/land-patterns :
...
+  chamfered-rect-pth-pad(C, drill-w, drill-h, pad-w, pad-h, pad-r)
+
+public defn chamfered-rect-smd-pad (w:Double, h:Double, r:Double) :
+  generic-pad(chamfered(w, h, r))
+
+;==============================================================================
+;========================= Land Patterns ======================================
+;==============================================================================
+; Create a non-plated hole of arbitrary shape for hole and mask
+public pcb-landpattern non-plated-hole-landpattern (hole:Shape, mask:Shape):
```
CuppoJava, line 299: This is a good helper definition.

utils/land-patterns.stanza
```diff
@@ -5,281 +5,406 @@ defpackage ocdb/land-patterns :
...
+
+;==============================================================================
+;========================== Land Pattern Generators ===========================
+;==============================================================================
+
+;==============================================================================
+;========================== Pose Arrangement Helpers ==========================
+;==============================================================================
+; PinPadMapping is a not-yet-placed pad, with assigned name, pad, and pose
+public pcb-struct ocdb/land-patterns/PinPadMapping:
```
CuppoJava, line 327: The reified type `PinPadMapping` and helper `make-generic-landpattern` is too sophisticated to expose publicly. It seems like they are for internal use, so please make them private.

CuppoJava (changes requested):
Good job on this PR. I want to get this fixed up and merged, and I left some important comments to address. The underlying logic looks solid to me, but I've been pickier than normal about coding style and API design because OCDB is public-facing and represents the code that we expect users to read and write.

utils/connects.stanza
m-hilgendorf: The reason for the use of this helper function was to check that all bundles matched their expected bundles and report them altogether, as opposed to checking one at a time and failing at the first change.

With the change to `ensure-bundle-port!` I don't see a clean way of representing this - would `error-buffer` be appropriate, or is this a pattern to avoid in general?

utils/land-patterns.stanza
m-hilgendorf: That is a major blocker, since we cannot rebuild a test release yet. OCDB has to be tested against the most recent stable release and cannot refer to code that isn't in the stable client that we test against.

utils/land-patterns.stanza
m-hilgendorf: I'm still unclear how to use those functions and why they are not in ocdb. I need explicit control over the `x,y` positions of every pad starting from an origin point.

utils/land-patterns.stanza
m-hilgendorf: My confusion is how to use these functions. I need total control over the placement of every pad, but the `-locs` family of functions don't appear to give that to me. They are also undocumented (both internally and in ocdb).

utils/land-patterns.stanza
m-hilgendorf: At the time this was written we didn't have pad introspection and this was not possible. I can refactor but it will take a lot of time, the switch to `generic-pad` was the most important bit of this PR.

utils/connects.stanza
CuppoJava: I see. You can do this also:
```
ensure-bundle-port!([x, y], [i2c, i2c])
```
The purpose is to reduce the complexity of the `ensure-bundles!` helper.

utils/land-patterns.stanza
CuppoJava: Okay. Let's merge it and stage it then. 

This part of CI needs to be figured out.

utils/land-patterns.stanza
CuppoJava: There is some old documentation in `notes/macro-doc.txt`, please take a look at that. 

Can you ask me some specific questions about what knobs you need?

components/texas-instruments/ISO1540.stanza
d-haldane: I reviewed this and like it. Actual ports on components are very handy.

CuppoJava: Perfect. Thanks for confirming.

PR #53, opened 2021-06-08
Box symbol should list pins in the order they appear in the pin-properties table.

CuppoJava:
Sort all pins by the order in which they are referred to in the pin-properties table before creating the box symbol.


ariel-hi (dismissed):
Reviewing in tandem with [JITPCB PR 53](https://github.com/JITx-Inc/jitpcb/pull/536).

When a box symbol is made, pins are sorted inside the pcb-component according to their `pin-properties-row-index` property. Specialized comparator.

Nice.

utils/box-symbol.stanza
```diff
@@ -244,6 +244,22 @@ public defn make-box-symbol (banks:Seqable<BoxBank>) :
 ;=======================================================<doc>
 
 public defn make-box-symbol () :
+  ;Sort the given pins by their .pin-properties-row-index field.
+  defn sort-by-row-index (pins:Tuple<JITXObject>) -> Tuple<JITXObject> :
+    ;Compare pin[i0] versus pin[i1].
+    ;Indices are used, instead of the pin directly, in order to perform a
+    ;stable sort.
+    ;All pins without a .pin-properties-row-index is put in the beginning.
```
bpellenard, line 252: are put

utils/box-symbol.stanza
```diff
@@ -244,6 +244,22 @@ public defn make-box-symbol (banks:Seqable<BoxBank>) :
...
+    defn compare-index (i0:Int, i1:Int) :
+      val a = pins[i0]
+      val b = pins[i1]
+      match(value?(property?(a.pin-properties-row-index)), value?(property?(b.pin-properties-row-index))) :
+        (ai:Int, bi:Int) : compare(ai, bi)
+        (ai:Int, bi) : 1
+        (ai, bi:Int) : -1
+        (ai, bi) : compare(i0,i1)
+    val indices = qsort(0 to length(pins), compare-index)
```
bpellenard, line 261: `qsort` or `qsort!`

utils/box-symbol.stanza
```diff
@@ -244,6 +244,22 @@ public defn make-box-symbol (banks:Seqable<BoxBank>) :
...
 
 public defn make-box-symbol () :
+  ;Sort the given pins by their .pin-properties-row-index field.
+  defn sort-by-row-index (pins:Tuple<JITXObject>) -> Tuple<JITXObject> :
+    ;Compare pin[i0] versus pin[i1].
+    ;Indices are used, instead of the pin directly, in order to perform a
+    ;stable sort.
+    ;All pins without a .pin-properties-row-index is put in the beginning.
+    defn compare-index (i0:Int, i1:Int) :
+      val a = pins[i0]
```
bpellenard, line 254: `val a = value?(property?(pins[i0].pin-properties-row-index))`

utils/box-symbol.stanza
ariel-hi: `qsort!` modifies the passed-in collection.
`qsort` returns a new collection.

utils/box-symbol.stanza
```diff
@@ -244,6 +244,22 @@ public defn make-box-symbol (banks:Seqable<BoxBank>) :
...
+      val b = pins[i1]
+      match(value?(property?(a.pin-properties-row-index)), value?(property?(b.pin-properties-row-index))) :
```
bpellenard, line 256:
```
val a = value?(property?(pins[i0].pin-properties-row-index))
val b = value?(property?(pins[i1].pin-properties-row-index))`
match(a, b) :
```

utils/box-symbol.stanza
CuppoJava: Thanks!

utils/box-symbol.stanza
CuppoJava: Thanks. That's cleaner.

bpellenard (approved):

utils/box-symbol.stanza
```diff
@@ -244,6 +244,24 @@ public defn make-box-symbol (banks:Seqable<BoxBank>) :
...
 public defn make-box-symbol () :
+  ;Sort the given pins by their .pin-properties-row-index field.
+  defn sort-by-row-index (pins:Tuple<JITXObject>) -> Tuple<JITXObject> :
+    ;Compare pin[i0] versus pin[i1].
+    ;Indices are used, instead of the pin directly, in order to perform a
+    ;stable sort.
+    ;All pins without a .pin-properties-row-index are put in the beginning.
+    defn compare-index (i0:Int, i1:Int) :
+      val a = pins[i0]
+      val b = pins[i1]
```
bpellenard, line 255: minor but you does not really need these two lines

CuppoJava: You're right. I would like it to be inlined too.

bpellenard (approved):

PR #54, opened 2021-06-10
Add helpers for courtyard and orientation markers

m-hilgendorf:
Two new API methods are added to utils/landpatterns:

`courtyard` computes a courtyard for a land pattern:

```
; Create a best guess of the courtyard using the existing artwork
public defn courtyard ()

; Create the courtyard of a component given its package dimensions. 
; The existing artwork of the landpattern will also be used to determine courtyard boundaries.
public defn courtyard (package-dims:Dims)
```

`orientation-marker` places a marker on the silkscreen relative to a pad to indicate the orientation of the component when placed:

```
public defn orientation-marker (
  pad-1:Ref,            ; the reference to the first pad of the landpattern (p[1], A[1], etc) 
  angle:Double|Anchor,  ; the angle or anchor relative to the first pad to draw the marker
  distance:Double,      ; the distance from the first pad to draw the marker
  shape:Shape           ; the shape of the marker itself
)   
```

Usage:

```
pcb-landpattern DFN-48:
  make-dfn-landpattern(48, 1.27, [0.3, 07])
  courtyard()
  orientation-marker(IndexRef(Ref("p"), 1), W, 1.27, Circle(0.5))
```


utils/land-patterns.stanza
```diff
@@ -23,6 +23,94 @@ public defn ref-label (x:Double, y:Double) :
...
+          switch(anchor):
+            C:  0.0  * PI
+            NE: 0.25 * PI
+            N:  0.5  * PI
+            NW: 0.75 * PI
+            W:  1.0  * PI
+            SW: 1.25 * PI
+            S:  1.5  * PI
+            SE: 1.75 * PI
+            E:  0.0
```
bpellenard, line 46: Same value than for `C`. Is it ok?

m-hilgendorf: Yes. `C` is an outlier, but I think it's better to be weird than to crash

utils/land-patterns.stanza
```diff
@@ -23,6 +23,94 @@ public defn ref-label (x:Double, y:Double) :
...
+        (s:Shape):
+          val dims = bounds(s)
+          layer(Courtyard(Top)) = Rectangle(x(dims), y(dims))
+        (_:False):
+          false
+
+public defn courtyard ():
+  courtyard(Dims(0.0, 0.0))
+
+; Create a capsule with normalized rotation
```
bpellenard, line 108: My original comment would have been the following: `normalize` is used in the reverse way in esir. For example, if you normalize an axis-aligned capsule, it will give you a capsule with an angle 0. 
But I am confused. First case the capsule is longer on the x-axis (width larger than height). Second case the capsule is longer on the y-axis. Is it on purpose?

utils/land-patterns.stanza
bpellenard: Yes, it is always tricky to handle it properly

utils/land-patterns.stanza
m-hilgendorf: Oh good catch. This should be removed from this PR, it's an abstraction of other code that's already in this file that was reused in a few places.

utils/land-patterns.stanza
```diff
@@ -23,6 +23,89 @@ public defn ref-label (x:Double, y:Double) :
...
+        for [ls, sh] in data seq?:
+          One(sh) when ls == side else None()
+      Union(shapes) when not empty?(shapes) else false
+    
+    for side in [Top, Bottom] do:
+      match(make-union?(side)):
+        (s:Shape):
+          val dims = bounds(s)
+          layer(Courtyard(Top)) = Rectangle(x(dims), y(dims))
+        (_:False):
```
bpellenard, line 102: `(_)`?

m-hilgendorf: I'm not familiar with that syntax, but this should only be `Shape|False`.

utils/land-patterns.stanza
```diff
@@ -23,6 +23,89 @@ public defn ref-label (x:Double, y:Double) :
...
+        val pad-shape_ = pad-shape(get-pad(pad_))
+        val pad-copper = 
+          if pad-type(get-pad(pad_)) == SMD:
+            [[side(pad_), pad-shape_]]
+          else:
+            [[Top, pad-shape_], [Bottom, pad-shape_]]
+        cat-all([pad-copper, layer-filter(layers(get-pad(pad_)))])
+
+    val data = to-tuple(cat-all([lp-data, pad-data, pkg-data]))
+    defn make-union? (side:Side) -> False|Shape:
```
bpellenard, line 91: I added `EmptyShape` if it helps here

utils/land-patterns.stanza
```diff
@@ -23,6 +23,89 @@ public defn ref-label (x:Double, y:Double) :
...
+    
+    for side in [Top, Bottom] do:
+      match(make-union?(side)):
+        (s:Shape):
+          val dims = bounds(s)
+          layer(Courtyard(Top)) = Rectangle(x(dims), y(dims))
+        (_:False):
+          false
+
+public defn courtyard ():
```
bpellenard, line 105: `make-default-coutyard`?

utils/land-patterns.stanza
m-hilgendorf: Does `Union(...)` return EmptyShape?

bpellenard:
@m-hilgendorf I have seen the same change in another PR. Which one do you want to keep?

bpellenard (approved):

m-hilgendorf:
@bpellenard which change is duplicated?

CuppoJava:
I don't quite understand the use case of this API. Isn't the courtyard directly determined by the landing pattern type? I.e. Shouldn't the `make-dfn-landpattern` generator know everything it is about to do, and therefore know precisely what the bounds of the courtyard need to be, and where the orientation marker needs to be?

CuppoJava (dismissed):
The code looks good.

m-hilgendorf:
This is a helper function intended to be called by generators like `make-qfn-pkg`.  A generator cannot know its bounds from its arguments, they must be calculated based on what pads are generated and the package size.

CuppoJava:
The part that I don't understand is: since the generator determines what pads to generate based upon the arguments, wouldn't it also know the appropriate courtyard based upon the same arguments? I think I'm missing something basic.

m-hilgendorf:
Yup! That's sort of what this helper is there for - first the generators place all their pads. Next, the generator has to figure out the bounding box of all the layers it has placed, and its package (which may be larger or smaller than the pads. The courtyard is the bounding box of all these layers. 

There's a bit of a discrepancy in what "courtyard" means according to some of the docs. The simplest definition is, "the bounding box of all the layers of the component except the silkscreen." That is what this helper is there to do.

PR #56, opened 2021-06-11
Improve `connects()` helpers in connects.stanza

m-hilgendorf:
- refactored ISO1540 to contain i2c ports explicitly
- refactored connects.stanza to check port bundles and print all mismatches 
- added unit tests for connects

### DIFF
--- a/tests/test-connects.stanza
+++ b/tests/test-connects.stanza
@@ -1,0 +1,75 @@
+#use-added-syntax(jitx, tests)
+defpackage ocdb/tests/connects: 
+  import core
+  import collections
+  import jitx
+  import jitx/commands
+  import ocdb/connects
+  import ocdb/bundles
+
+defn expect-fatal (body:() -> ?):
+  var caught? = false
+  execute-with-error-handler(
+    body,
+    fn (): caught? = true
+  )
+  #EXPECT(caught?)
+
+deftest(ocdb, connects) test-connect-i2c-isolation:
+  pcb-module M1:
+    port p: i2c[[1, 2]]
+    #EXPECT(connect-i2c-isolation(p[1], p[2]) is-not False)
+  
+  pcb-module M2:
+    port p: i2c[[1, 2]]
+    #EXPECT(connect-i2c-isolation(p[1], p[2], M1) is-not False)
+
+  pcb-module M3:
+    port x: i2c
+    port y: gpio
+    var caught? = false
+    within expect-fatal():
+      connect-i2c-isolation(x, y)
+     
+  for mod in [M1, M2, M3] do:
+    print-def(mod)
+
+deftest(ocdb, connects) test-connect-phy:
+  pcb-module M1:
+    port x: rgmii
+    port y: ethernet-1000
+    #EXPECT(connect-phy(x, y) is-not False)
+  pcb-module M2:
+    port x: rgmii
+    port y: gpio
+    within expect-fatal():
+      connect-phy(x, y)
+  pcb-module M3:
+    port x: gpio
+    port y: ethernet-1000
+    within expect-fatal():
+      connect-phy(x, y)
+  for mod in [M1, M2, M3] do:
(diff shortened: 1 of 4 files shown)


tests/test-connects.stanza
```diff
@@ -0,0 +1,75 @@
...
+
+defn expect-fatal (body:() -> ?):
+  var caught? = false
+  execute-with-error-handler(
+    body,
+    fn (): caught? = true
+  )
+  #EXPECT(caught?)
+
+deftest(ocdb, connects) test-connect-i2c-isolation:
```
bpellenard, line 18: Nice!

utils/connects.stanza
```diff
@@ -12,41 +12,89 @@ defpackage ocdb/connects :
...
+        (b?:Bundle):
+          if b? != b:
+            One([p, b, b?])  ; case 1: ports exist but do not match
+          else:
+            None()           ; case 2: ports exist and match, we're good to go
+        (_:?):
+          One([p, b, false]) ; case 3: port is not a bundle
+  
+  ; error if not empty
+  if not empty?(errors):
```
bpellenard, line 37: I would reverse it: `if empty?(errors) : else :`

bpellenard (approved):

utils/connects.stanza
```diff
@@ -12,41 +12,89 @@ defpackage ocdb/connects :
...
+    val buf = StringBuffer()
+    println(buf, "Could not %_. Bundles did not match." % [action])
+    for [p, b, b?] in errors do:
+      println(IndentedStream(buf), 
+        "Expected %_ for port %_. Found %_." % [
+          name(b), ref(p), 
+          name(b? as Bundle) when b? is-not False else false
+        ]
+      )
+    fatal(to-string(buf)) ; todo error or exception here?
```
ariel-hi, line 47: I would make a public `BundleNotFound` (or a more apt name) exception in connections.stanza
This should be an exception since it occurs when a designer provides a bad object, and not because of a connections implementation error.

utils/connects.stanza
```diff
@@ -12,41 +12,89 @@ defpackage ocdb/connects :
...
+  
+  ; error if not empty
+  if not empty?(errors):
+    val buf = StringBuffer()
+    println(buf, "Could not %_. Bundles did not match." % [action])
+    for [p, b, b?] in errors do:
+      println(IndentedStream(buf), 
+        "Expected %_ for port %_. Found %_." % [
+          name(b), ref(p), 
+          name(b? as Bundle) when b? is-not False else false
```
ariel-hi, line 44: `else "no bundle"` instead of `else false` ?
Also, `else false` wouldn't be needed since false is default.

utils/connects.stanza
```diff
@@ -12,41 +12,89 @@ defpackage ocdb/connects :
...
+            One([p, b, b?])  ; case 1: ports exist but do not match
+          else:
+            None()           ; case 2: ports exist and match, we're good to go
+        (_:?):
+          One([p, b, false]) ; case 3: port is not a bundle
+  
+  ; error if not empty
+  if not empty?(errors):
+    val buf = StringBuffer()
+    println(buf, "Could not %_. Bundles did not match." % [action])
```
ariel-hi, line 39: Nice use of action :)

ariel-hi (commented):
Big fan of the connects cleanup! Comments on error thrown.

utils/connects.stanza
m-hilgendorf: I feel like that falls under "blame the caller." The error handling in ocdb is already up in the air - I don't think we throw exceptions anywhere internally in this library. But I'm not sure what our error handling story should look like

ariel-hi: Agreed. Either we have an `ocdb/errors.stanza` or inline the Exception to this file.

bpellenard (approved):

PR #51, opened 2021-06-03
A large PR to review before merging into the pre-merge branch for the other PR into dev

d-haldane:

PR #61, opened 2021-06-15
STM importer helpers

m-hilgendorf:


m-hilgendorf:
Takes over for #51

components/abracon/AMPM.stanza
```diff
@@ -50,36 +53,8 @@ public pcb-component component (frequency:Double):
     50.0e6   : ["AMPMAFB-50.0000T"  25.0e-6  VFLGA-2 ]    
     8.0e6    : ["AMPMEFB-8.0000T"   25.0e-6  VDFN    ]  
     80.0e6   : ["AMPMAFB-80.0000T"  25.0e-6  VFLGA-2 ]    
-
-    ; 10.0e6   : ["AMPMGFB-10.0000T"  25.0e-6  VFLGA-16]
```
d-haldane, line 54: These lines, although commented out, still contain useful information. Can we keep them?

m-hilgendorf: How is the information intended to be used? I'm hesitant to keep large chunks of code that isn't functional. We can always add it back later.

d-haldane (approved):

components/american-opto/L196L.stanza
```diff
@@ -0,0 +1,39 @@
...
+  import ocdb/defaults
+
+  import ocdb/land-patterns
+  import ocdb/symbols
+  import ocdb/box-symbol
+  import ocdb/bundles
+  import ocdb/generator-utils
+  import ocdb/generic-components
+
+;<test>
```
bpellenard, line 17: @m-hilgendorf , in another PR, @PhilippeFerreiraDeSousa wants to remove these comments

m-hilgendorf: These are still necessary, and the tests will break in this branch without the annotations.

PhilippeFerreiraDeSousa: The other PR changes the fallback behaviour if there is no annotation of argument values:
- tries to guess the argument value of the macro on its own
- else skip it

components/abracon/AMPM.stanza
d-haldane: It's used to model the rest of the devices in this family. The component would get an extra argument for package, and a few more landpattern options would be added. I like having breadcrumbs that make it easy to keep adding to ocdb but if you find it disruptive, it's ok to delete it.

PR #65, opened 2021-07-04
JITX-636: Refactored landpatterns

bpellenard:


utils/land-patterns.stanza
```diff
@@ -277,19 +285,22 @@ public defn dshape-pth-pad (drill-w:Double, drill-h:Double, pad-w:Double, pad-h:
...
-  layer(SolderMask(Top)) = Circle(C, mask-r)
-  layer(SolderMask(Bottom)) = Circle(C, mask-r)
-  layer(Courtyard(Top)) = Rectangle((drill-r * 2.0), (drill-r * 2.0))
-  layer(Courtyard(Top)) = Rectangle((drill-r * 2.0), (drill-r * 2.0))
+  val circle = Circle(C, mask-r)
+  layer(SolderMask(Top)) = circle
+  layer(SolderMask(Bottom)) = circle
+  val rectangle = Rectangle((drill-r * 2.0), (drill-r * 2.0))
+  layer(Courtyard(Top)) = rectangle
+  layer(Courtyard(Top)) = rectangle
```
bpellenard, line 303: should not it be `layer(Courtyard(Bottom)) = rectangle`

PhilippeFerreiraDeSousa: Oh, we should probably throw an esir error when somebody try to define the same layer twice

bpellenard: Is it easy to detect?

m-hilgendorf:
This PR duplicates a lot of the effort in #52 which changes how additional layers are generated.

utils/land-patterns.stanza
m-hilgendorf: This is not an error in general, you may place different shapes on the same layer.

bpellenard:
Ok, so I am going to close it

bpellenard:
Already done in https://github.com/JITx-Inc/open-components-database/pull/52

PR #68, opened 2021-07-08
Partial set of new bundles for this week's sprint.

d-haldane:


RajeshJITX (approved):
Looks good. I will change the scripts that parse XML to use these new names for SPI (cs, sdi, sdo) instead of (ss, miso, mosi), jtag etc.

RajeshJITX (dismissed):
change uart-with to usart

utils/bundles.stanza
```diff
@@ -29,39 +29,35 @@ public pcb-bundle power :
 public pcb-bundle fpga-diff : 
   diff-pair-pins()
 
-public pcb-bundle jtag :
+public pcb-enum ocdb/bundles/JTAGPins :
+  TRSTN
```
m-hilgendorf, line 33: Since enum variants are global, they should be prefixed. The convention we use is `NameOfEnumNameOfVariant`, in this case:

```public pcb-enum ocdb/bundles/JTAGPins:
  JTAGTRSTN
```

That said we should probably notuse single variant enums, a boolean as a generator argument would work fine. Is this enum going to be expanded later?

utils/bundles.stanza
```diff
@@ -78,35 +74,75 @@ public pcb-bundle swd:
   pin swdio
   pin swdclk
 
-public pcb-bundle uart :
+public pcb-enum ocdb/bundles/UARTPins :
```
m-hilgendorf, line 77: same as above, add a `UartDTR` or similar

utils/bundles.stanza
```diff
@@ -29,39 +29,35 @@ public pcb-bundle power :
 public pcb-bundle fpga-diff : 
   diff-pair-pins()
 
-public pcb-bundle jtag :
+public pcb-enum ocdb/bundles/JTAGPins :
+  TRSTN
+
+public pcb-bundle jtag (pins:Tuple<ocdb/bundles/JTAGPins>):
```
m-hilgendorf, line 35: I believe you only need the fully qualified name in the type definition, this should be ok:

```
public pcb-bundle jtag (pins:Tuple<JTAGPin>):
```

utils/bundles.stanza
```diff
@@ -78,35 +74,75 @@ public pcb-bundle swd:
...
-public pcb-bundle uart :
+public pcb-enum ocdb/bundles/UARTPins :
+  DTR
+  CTS
+  DCD
+  RI
+  DST
+  RTS
+
+pcb-bundle uart (pins:Tuple<ocdb/bundles/UARTPins>) :
```
m-hilgendorf, line 85: This definition of `uart` is identifical to `usart`. Do we require them to be unique?

utils/bundles.stanza
```diff
@@ -78,35 +74,75 @@ public pcb-bundle swd:
...
+      RI  : make-pin(`ri)
+      DST : make-pin(`dst)
+      RTS : make-pin(`rts)
+
+public pcb-bundle uart ():
   pin rx
   pin tx
 
-public pcb-bundle uart-with (ctrl-pins:Tuple<Symbol>) :
+pcb-bundle usart (pins:Tuple<ocdb/bundles/UARTPins>) :
```
m-hilgendorf, line 101: This change breaks `scripts/cubemx-importer.stanza` which relied on `uart-with`

utils/bundles.stanza
```diff
@@ -78,35 +74,75 @@ public pcb-bundle swd:
...
+  for p in pins do :
+    switch(p) :
+      DTR : make-pin(`dtr)
+      CTS : make-pin(`cts)
+      DCD : make-pin(`dcd)
+      RI  : make-pin(`ri)
+      DST : make-pin(`dst)
+      RTS : make-pin(`rts)
+
+public pcb-bundle usart ():
```
m-hilgendorf, line 113: We should make sure to reuse the generator here 

```
public defn usart ():
  usart([])
```

utils/bundles.stanza
```diff
@@ -78,35 +74,75 @@ public pcb-bundle swd:
...
+  for p in pins do :
+    switch(p) :
+      SDI : make-pin(`sdi)
+      SDO : make-pin(`sdo)
+      CS  : make-pin(`cs)
+      CIPO  : make-pin(`cipo)
+      COPI  : make-pin(`copi)
+
+; Most common spi bundle
+public pcb-bundle spi ():
```
m-hilgendorf, line 136: same as above

```
public defn pcb-bundle spi ():
  spi([])
```

m-hilgendorf (changes requested):
Love the change to `pcb-enum`. This is a big breaking change, however and the tests should report most of the failures, but the cubemx importer needs to be updated to use the enum variants, new bundle signatures, and enum variants. 

@CuppoJava, do we have a preference for style on pcb-enum variants? `JTAGTRSTN` looks pretty difficult to read.

CuppoJava:
It's slightly ugly, but the prefix is important. Normally, we use CamelCase for enum names, but in this case, since everything is a capital letter, let's add a hyphen?
```
JTAG-TRSTN
```

utils/bundles.stanza
d-haldane: Yes - software support is very different

utils/bundles.stanza
```diff
@@ -82,35 +75,76 @@ public pcb-bundle swd:
...
+  SPI-CS
+  SPI-CIPO
+  SPI-COPI
+  SPI-SDIO
+
+public pcb-bundle spi  (pins:Tuple<ocdb/bundles/SPIPins>):
+  pin sck
+  for p in pins do :
+    switch(p) :
+      SPI-SDI : make-pin(`sdi)
```
m-hilgendorf, line 134: :+1:

m-hilgendorf (commented):
lgtm, just needs to pass tests and we can merge

utils/bundles.stanza
```diff
@@ -82,35 +75,76 @@ public pcb-bundle swd:
...
+      UART-DTR : make-pin(`dtr)
+      UART-CTS : make-pin(`cts)
+      UART-DCD : make-pin(`dcd)
+      UART-RI  : make-pin(`ri)
+      UART-DST : make-pin(`dst)
+      UART-RTS : make-pin(`rts)
+      UART-CK : make-pin(`ck)
+      UART-DE : make-pin(`de)
+
+public pcb-bundle uart ():
```
m-hilgendorf, line 102: needs to be `defn`

utils/bundles.stanza
m-hilgendorf: needs to be `defn` not `pcb-bundle` here too

utils/bundles.stanza
```diff
@@ -82,35 +75,76 @@ public pcb-bundle swd:
...
+  for p in pins do :
+    switch(p) :
+      SPI-SDI : make-pin(`sdi)
+      SPI-SDO : make-pin(`sdo)
+      SPI-CS  : make-pin(`cs)
+      SPI-CIPO  : make-pin(`cipo)
+      SPI-COPI  : make-pin(`copi)
+
+; Most common spi bundle
+public pcb-bundle spi ():
```
m-hilgendorf, line 141: `public defn spi ()`

m-hilgendorf (changes requested):

m-hilgendorf (approved):
renames look good and tests pass, thanks @RajeshJITX!

PR #72, opened 2021-07-10
Compile error found during compiling of jitx-client

RajeshJITX:
Currently, not all OCDB files are compiled when changes are made to OCDB. Some errors are found during compilation of jitx-client. This PR fixes those errors

### DIFF
--- a/components/texas-instruments/CC2640.stanza
+++ b/components/texas-instruments/CC2640.stanza
@@ -94,16 +94,16 @@ public pcb-component component :
 
   supports uart():
     require pins:io-pin[2]
-    uart.rx    =>    pins[0].p
-    uart.tx    =>    pins[1].p
+    uart().rx    =>    pins[0].p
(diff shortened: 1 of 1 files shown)


bpellenard (approved):

PhilippeFerreiraDeSousa:
This file is compiled in ocdb CI but the macro did not evaluate correctly and nobody knew how to fix it so I was told to skip the test for it, cf `;<test>skip<test>` line 26.

PhilippeFerreiraDeSousa:
@RajeshJITX Actually you shoud have removed the `;<test>skip<test>` before merging

RajeshJITX:
I did not change that part. This kind of change mixes intent of the ticket. Will create a new ticket to skip testing or code cleanup which is not relevant

PhilippeFerreiraDeSousa:
It is not very important, it is fine. But it it is precisely related as you fixed the test that we skipped because it was failing.

PR #78, opened 2021-07-14
JITX-647: Added landpatterns utility functions for DShape, GeneralRoundedRectangle and GeneralChamferedRectangle

bpellenard:
…

### DIFF
--- a/utils/land-patterns.stanza
+++ b/utils/land-patterns.stanza
@@ -90,26 +90,6 @@ defn chamfered (anchor:Anchor, width:Double, height:Double, r:Double):
   val [length, width_, rot] = normalize-rotation(width, height)
   normalize-loc(ChamferedRectangle(length, width_, r), anchor, rot)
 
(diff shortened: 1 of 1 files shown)


m-hilgendorf (approved):
Looks great! The only things it the usage of single-character arguments but that is consistent with the rest of land-patterns.stanza.

CuppoJava (dismissed):
This looks good to me. 

Can we make a JIRA ticket or TODO comment that asks us to revisit why there is this extra level of indirection in shape-creation? It seems `general-round-rect` and `GeneralRoundedRectangle` are identical, so I'm not sure what the extra code is for.

bpellenard:
I removed the useless additional layer. It is useless since there is no need for normalization for `DShape`, `GeneralRoundedRectangle` and `GeneralChamferedRectangle`

CuppoJava (approved):
Thanks! Looks good.

PR #82, opened 2021-07-19
Fixed offset/expand computation

m-hilgendorf:
Fixes the computation of `expansion` for a given solder mask layer on Capsule, DShape, and ChamferedRectangle pads. 

Before:
Screenshot from 2021-07-19 13-26-20

After:

Screenshot from 2021-07-19 13-25-42

`tests/test-pads.stanza` is a test harness for reproducing the results (Thanks, @ariel-hi!)

### DIFF
--- a/tests/test-pads.stanza
+++ b/tests/test-pads.stanza
@@ -1,0 +1,59 @@
+#use-added-syntax(jitx)
+defpackage pads :
+  import core
+  import collections
+  import math
+  import jitx
+  import jitx/commands
+  import ocdb/defaults
+
+  import ocdb/symbols
+  import ocdb/box-symbol
+  import ocdb/land-patterns
+  import ocdb/bundles
+  import ocdb/generator-utils
+  import ocdb/generic-components
+
+pcb-landpattern lp (n:Int) :
+  for i in 0 to n by 4 do :
+    val d = to-double(i) * 2.0
+    val drill = 0.5
+    val x = 1.0
+    val y = 2.0
+    pad p[i] : pth-pad(drill / 2., x / 2.) at loc(d, d)
+    pad p[i + 1] : oval-pth-pad(drill, x, y) at loc(d + 2.0, d + 2.0)
+    pad p[i + 2] : rect-pth-pad(drill, x, y) at loc(d + 4.0, d + 4.0)
+    pad p[i + 3] : round-rect-pth-pad(drill, x, y, 0.25) at loc(d + 6.0, d + 6.0)
+    pad p[i + 4] : dshape-pth-pad(drill, x, y, y, 0.5) at loc(d + 8.0, d + 8.0)
+    pad p[i + 5] : chamfered-rect-pth-pad(drill, x, y, y, 0.5) at loc(d + 10.0, d + 10.0)
+  ref-label()
+
+pcb-symbol sym (n:Int) :
+  for i in 0 to n do :
+    val d = to-double(i) * 2.54
+    pin p[i] at Point(d, d)
+
+pcb-component ldo-9000-sym (n:Int) :
+  name = "AP2112"
+  manufacturer = "Diodes Incorporated"
+  description = "600-mA, Low-Dropout Regulator"
+  mpn = "AP2112K-3.3TRG1"
+  reference-prefix = "A"
+
+  port p : pin[n]
+
+  val sym = sym(n)
+  val lp = lp(n)
+  symbol = sym(for i in 0 to n do : p[i] => sym.p[i])
+  landpattern = lp(for i in 0 to n do : p[i] => lp.p[i])
+
+pcb-module ldo :
+  val num-pins = 4
+  public inst l : ldo-9000-sym(num-pins)[4]
+
+set-export-backend(`kicad)
+make-default-board(ldo, 4, Rectangle(10.0, 10.0))
+
+view-board()
+;view-schematic()
+;export-cad()
--- a/tests/test-pads.stanza
+++ b/tests/test-pads.stanza
@@ -15,7 +15,7 @@ defpackage pads :
   import ocdb/generic-components
 
 pcb-landpattern lp (n:Int) :
-  for i in 0 to n by 4 do :
+  for i in 0 to n by 6 do :
     val d = to-double(i) * 2.0
     val drill = 0.5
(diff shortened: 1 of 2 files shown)


utils/land-patterns.stanza
```diff
@@ -147,11 +147,13 @@ public defn apply-soldermask (mask-shape:Shape):
 public defn apply-soldermask (amount:Double, expand?:True|False):
   val mask-shape = 
     match(pad-shape(self)):
-      (s:Rectangle|Circle|RoundedRectangle|Segment):
+      (s:Rectangle|Circle|RoundedRectangle|Segment|Capsule):
```
m-hilgendorf, line 150: `offset` is implemented for `Capsule`, so we can include it here.

tests/test-pads.stanza
```diff
@@ -0,0 +1,59 @@
...
+
+  import ocdb/symbols
+  import ocdb/box-symbol
+  import ocdb/land-patterns
+  import ocdb/bundles
+  import ocdb/generator-utils
+  import ocdb/generic-components
+
+pcb-landpattern lp (n:Int) :
+  for i in 0 to n by 4 do :
```
ariel-hi, line 18: Can you change the 4 here to a 6 to reflect the number of pads created per loop? This design will break if `num-pins` on line 51 is at least 5.

tests/test-pads.stanza
```diff
@@ -0,0 +1,59 @@
...
+
+  port p : pin[n]
+
+  val sym = sym(n)
+  val lp = lp(n)
+  symbol = sym(for i in 0 to n do : p[i] => sym.p[i])
+  landpattern = lp(for i in 0 to n do : p[i] => lp.p[i])
+
+pcb-module ldo :
+  val num-pins = 4
```
ariel-hi, line 51: And this to a 6.

utils/land-patterns.stanza
```diff
@@ -147,11 +147,13 @@ public defn apply-soldermask (mask-shape:Shape):
...
         val func = expand when expand? else offset
         func(s, amount)
-      (s:ChamferedRectangle|DShape|Capsule):
-        expand(s, amount)
+      (s:ChamferedRectangle|DShape):
+        val d = dims(s)
+        val expansion = 1.0 + amount / max(x(d), y(d))
```
m-hilgendorf, line 155: the logic is as follows: 

`expand(shape, x)` will grow or shrink `shape` by the non-negative factor `x`.

The generator is passed `amount` as an offset, so we need to compute the most sane expansion factor to approximate. This is due to an internal limitation on the implementation of `offset` which does not cover DShape and ChamferedRectangle yet. 

To do this, we find the maximum of the x/y dimensions of the shape. Dividing the amount by this value gives us the expansion as a ratio, adding it to one gives us the overall expansion factor. `amount` may be negative, in which case the shape will shrink.

ariel-hi (approved):

tests/test-pads.stanza
```diff
@@ -0,0 +1,59 @@
...
+  import ocdb/symbols
+  import ocdb/box-symbol
+  import ocdb/land-patterns
+  import ocdb/bundles
+  import ocdb/generator-utils
+  import ocdb/generic-components
+
+pcb-landpattern lp (n:Int) :
+  for i in 0 to n by 6 do :
+    val d = to-double(i) * 2.0
```
bpellenard, line 19: `* 12.0` instead?
Have you tried with an input `n` larger than `6`?

utils/land-patterns.stanza
bpellenard: Why this time the expansion parameter has no dimension (ratio of two lengths)? (and for other cases `expand(s, amount)`)

tests/test-pads.stanza
m-hilgendorf: No. Why * 12?

utils/land-patterns.stanza
m-hilgendorf: Can you clarify what you mean by dimension?

tests/test-pads.stanza
bpellenard: My bad

utils/land-patterns.stanza
bpellenard: unit

m-hilgendorf: For `expand (x, factor)`, factor is unitless. It enlarges or shrinks by a scalar. For example:

```
val r1 = Rectangle(1.0, 1.0)
val r2 = expand(r1, 2.0)

r2 == Rectangle(2.0, 2.0)
```

bpellenard: The `amount` parameter name is kind of confusing

m-hilgendorf: I removed the option to use `expand` and renamed to `expansion-factor` in the code.

utils/land-patterns.stanza
CuppoJava: This behaves differently than before, and a big part of the original discussion was that the solder mask just keep the same shape as the original. So the solder mask for a Rectangle should still be just a simple Rectangle. Is this requirement still true?

utils/land-patterns.stanza
```diff
@@ -178,8 +180,8 @@ public defn soldermask-defined (p:Pad):
     pcb-pad SMD-p:
       name  = to-string("%_, SMD" % [name(p)])
       type  = pad-type(p)
-      shape = pad-shape(p)
-      apply-soldermask(0.85, true)
+      shape = offset(pad-shape(p), SOLDER-MASK-REGISTRATION)
```
CuppoJava, line 183: This fix is fine for now, but this utility is overly general and there are PRs coming that moves solder-mask-defined flag to within the individual landpattern generators.

utils/land-patterns.stanza
m-hilgendorf: We're using `offset` here as the default because it is better defined for what "solder mask registration" means (the distance away from the pad/mask shape that defines the pad).

utils/land-patterns.stanza
m-hilgendorf: This should probably be replaced once we have a general purpose substitution mechanism as well.

bpellenard (approved):

PR #85, opened 2021-07-20
Add timer support

RajeshJITX:
Generate timer support statements for timer pins that have format TIM12_CH3 only. Generated statements look like

```
    supports timer:
      timer.timer => self.PB[6]

    supports timer:
      timer.timer => self.PB[5]

```

### DIFF
--- a/scripts/cubemx-importer.stanza
+++ b/scripts/cubemx-importer.stanza
@@ -44,6 +44,10 @@ defstruct AdcOption:
 defstruct DacOption:
   dac:Stm32Pin
 
+; An option for Timer
+defstruct TimerOption:
+  timer:Stm32Pin
+
 ; An option for an SWD port
 defstruct SwdOption:
   swdio:Stm32Pin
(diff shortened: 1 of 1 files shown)


scripts/cubemx-importer.stanza
```diff
@@ -305,6 +314,20 @@ defn dac-pin? (s:String) -> True|False :
...
   to-tuple(seq(DacOption, pins-with-signal(pins, dac-pin?)))
 
+; Only generates support statements for timer pins that have
+; format TIM[0-9]+_CH[0-9]+ - example TIM3_CH1
+; It ignores other types (exhaustive list of formats)
+; TIM15_BK, TIM15_BKIN, TIM15_CH1N, TIM19_ETR, TIM1_BK2,
+; TIM1_BKIN2, TIM1_BKIN2_COMP1, TIM1_BKIN_COMP1, TIMX_IC1
+defn timer-pin? (s:String) -> True|False :
+  val b = not empty?(regex-match("^TIM[0-9]+_CH[0-9]+[^N]", s))
+  ;println("%_ : %_" % [s, b])
```
m-hilgendorf, line 324: we can use

```
not empty?(regex-match("^TIM[0-9]+_CH[0-9]+[^N]", s))
```

here w/o the print debugging

m-hilgendorf (changes requested):

scripts/cubemx-importer.stanza
RajeshJITX: I have tested this file on 6 files. I need to test them on another 1200+ files. So I would like to keep this debug statement until I test against more of the files

m-hilgendorf: Then we should wait until it's fully tested before merging, we really shouldn't be putting print debugging into production.

m-hilgendorf (approved):

d-haldane (approved):

PR #86, opened 2021-07-22
IPC-Compliant Two-pin landpattern generator

m-hilgendorf:
Example:

Screenshot from 2021-07-22 14-40-38

This PR adds a set of generators for creating IPC compliant two-pin landpatterns given the manufacturing and fabrication tolerances.


stanza.proj
```diff
@@ -2,11 +2,11 @@
 ; Land pattern and symbol utilities 
 ;====================================
 package ocdb/land-patterns defined-in "utils/land-patterns.stanza"
-package ocdb/land-patterns/ipc-helpers defined-in "utils/land-patterns/ipc-helpers.stanza"
-package ocdb/land-patterns/leads defined-in "utils.land-patterns/ipc-helpers.stanza"
+package ocdb/land-protrusions defined-in "utils/land-protrusions.stanza"
```
m-hilgendorf, line 5: rename old file to `land-protrusions`

tests/test-land-patterns.stanza
```diff
@@ -0,0 +1,80 @@
...
+pcb-component example-2:
+  pin a
+  pin c
+  reference-prefix = "C"
+  pcb-landpattern lp:
+    make-two-pin-chip-landpattern(
+      Toleranced(3.2, 0.2),
+      Toleranced(1.6, 0.2),
+      Toleranced(0.8, 0.3),
+      Toleranced(1.2, 0.1),
```
m-hilgendorf, line 60: This example has a polarity marker and override for lead-width that is smaller than the component-width.

utils/design-vars.stanza
```diff
@@ -18,11 +21,15 @@ public var DESIGN-QUANTITY = 100
 public var PREFERRED-MOUNTING = "smd" ; values in ["smd", "through-hole"]
 public var MIN-CAP-VOLTAGE = 10.0 ; Minimum voltage to allow for capacitors
 
-; For consolidation
-;public var PREFERRED-MANUFACTURERS = ["Yageo", "Panasonic Electronic Components", "Vishay Dale", "TE Connectivity Passive Product"]
+; Landpattern variables
 public var SOLDER-MASK-REGISTRATION: Double = 0.15
+public var PLACEMENT-TOLERANCE   = 0.0 ; "P" on page 13 IPC-7351B. FIXME goes in pcb-rules
```
m-hilgendorf, line 26: Add better defaults

utils/land-patterns.stanza
```diff
@@ -18,6 +20,62 @@ defn ensure-even-positive! (value: Int, name: String):
   if value < 2 or value % 2 == 1:
     fatal("%_ only supports a positive even number of pins." % [name])
 
+; find the angle of a pad away from the origin
+defn pad-angle (pad-ref:Ref) -> Double:
```
m-hilgendorf, line 24: remove, unnecessary

utils/land-patterns.stanza
```diff
@@ -93,6 +151,21 @@ defn chamfered (anchor:Anchor, width:Double, height:Double, r:Double):
 ;==============================================================================
 ;=========================== Geometric Helpers ================================
 ;==============================================================================
+; ====== Land pattern utilities ===============================================
+; Find the extent dimensions of a landpattern given all its layers and pads.
+public defn extent-dims () -> Dims:
```
m-hilgendorf, line 156: sanity check current usage and remove if we aren't using it anymore

utils/land-patterns.stanza
```diff
@@ -153,8 +226,8 @@ public defn apply-soldermask (amount:Double, expand?:True|False):
       (s:ChamferedRectangle|DShape|Capsule):
```
m-hilgendorf, line 226: big TODO here: replace with new expand-by-amount function that is not equivalent to `offset` or `expand` (preserve shape, but increase dimensions additively rather than multiplicatively)

utils/land-patterns.stanza
```diff
@@ -558,10 +634,16 @@ public defn make-ipc-two-pin-landpattern (part-name:String, courtyard?:True|Fals
         val nr = Z
         val nc = X
         [Z, G, X, nr, nc]
-    make-two-pin-landpattern(Z, G, X, nr, nc, courtyard?, polarized?)
+    if v1?:
```
m-hilgendorf, line 637: delete this check

utils/land-patterns.stanza
```diff
@@ -558,10 +634,16 @@ public defn make-ipc-two-pin-landpattern (part-name:String, courtyard?:True|Fals
...
+      make-two-pin-landpattern(Z, G, X, nr, nc, courtyard?, polarized?)
+    else:
+      make-two-pin-landpattern(Z, G, X, nr, nc, courtyard?, polarized?)
     external-names = ([part-name])
     ref-label()
 
+public defn make-ipc-two-pin-landpattern (part-name:String, courtyard?:True|False, polarized?:True|False):
```
m-hilgendorf, line 644: remove

utils/tolerance.stanza
```diff
@@ -0,0 +1,77 @@
+#use-added-syntax(jitx)
+defpackage ocdb/tolerance:
+  import core
+  import collections
+  import math
+
+; A generic value with some tolerance. May be offcenter (upper tolerance != lower tolerance)
+public pcb-struct ocdb/tolerance/Toleranced:
```
m-hilgendorf, line 8: Is `Toleranced` the appropriate name?

RajeshJITX (dismissed):

utils/tolerance.stanza
```diff
@@ -0,0 +1,77 @@
...
+; Return the upper tolerance, as a percentage
+public defn tol+% (t:Toleranced) -> Double:
+  plus(t) / nominal(t)
+
+; Return the lower tolerance, as a percentage
+public defn tol-% (t:Toleranced) -> Double:
+  minus(t) / nominal(t)
+
+; Return the max value
+public defn max (t:Toleranced) -> Double:
```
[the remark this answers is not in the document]
CuppoJava: Agreed.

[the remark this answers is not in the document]
m-hilgendorf: Any suggestions on an alternative name?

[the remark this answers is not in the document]
CuppoJava: `max-value` is one suggestion.

utils/land-patterns.stanza
```diff
@@ -911,22 +965,22 @@ public pcb-landpattern INDC1005X55 :
 
 public pcb-landpattern CAPT1005X55 :
   make-ipc-two-pin-landpattern("0402")
-  layer(Silkscreen("pol", Top)) = PlusSymbol(-1.7, 0.0)
+  draw-polarity-marker(IndexRef(Ref("p"), 1))
```
CuppoJava, line 968: After the change to `draw-polarity-marker` to take `JITXObject`, you can call it like this:

```
draw-polarity-marker(self.p[1])
```

CuppoJava: It seems that the `two-pin-landpattern` should directly generate the polarity marker itself instead of relying upon callers to do it. Especially since it already takes a `polarized?` argument.

utils/land-patterns.stanza
```diff
@@ -1034,3 +1088,155 @@ public pcb-landpattern smd-testpoint-pkg (d:Double) :
...
+
+public defn make-two-pin-chip-landpattern (length:Toleranced,
+                                           width:Toleranced,
+                                           lead-length:Toleranced,
+                                           lead-width?:False|Toleranced, ; in case different from width
+                                           fabrication-tolerance:Double,
+                                           placement-tolerance:Double
+                                           density-level:DensityLevel,
+                                           polarized?:True|False):
+                                          ;  solder-mask-defined?:True|False): ;
```
CuppoJava, line 1100: Remove commented-out code from public-facing OCDB.

utils/tolerance.stanza
```diff
@@ -0,0 +1,60 @@
...
+public pcb-struct ocdb/tolerance/Toleranced:
+  tol+:Double,
+  tol-:Double,
+  typ: Double,
+
+defmethod print (o:OutputStream, t:Toleranced):
+  println(o, "Toleranced(%_ <= %_ <= %_)" % [min-value(t), typ(t), max-value(t)])
+
+; Create a tolerance value with the same tolerance in both directions
+public defn Toleranced (value:Double, tolerance:Double) -> Toleranced:
```
CuppoJava, line 17: Use the same convention for all the `Toleranced` constructors. 

Either they are called `Toleranced` to follow standard Stanza convention, or they have their own short names (e.g. `tol%`).

CuppoJava (changes requested):
Requested some cleanups.

utils/land-patterns.stanza
```diff
@@ -18,6 +20,54 @@ defn ensure-even-positive! (value: Int, name: String):
   if value < 2 or value % 2 == 1:
     fatal("%_ only supports a positive even number of pins." % [name])
 
+; ====== Silk screen helpers ===================================================
+public defn draw-polarity-marker (pad-ref:Ref, part-dims?:Dims|False) :
```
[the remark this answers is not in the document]
m-hilgendorf: part-dims are the physical dimensions of the part (not any of the layers associated with the land pattern). We don't have a way of representing this generally, so I pass them in as an argument.

utils/land-patterns.stanza
m-hilgendorf: This code does not call the two pin landpattern generator (and doing so would break it, because callers of this generator don't use `a` and `c` as the pad names).

utils/land-patterns.stanza
```diff
@@ -201,6 +252,9 @@ public pcb-pad smd-pad (s:Shape):
   name  = "SMD Pad"
   type  = SMD
   shape = s
+  ; TODO: add the soldermask manually here with offset so it's
```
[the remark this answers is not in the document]
m-hilgendorf: Removed

utils/land-patterns.stanza
m-hilgendorf: Added the changes

utils/land-patterns.stanza
m-hilgendorf: Done

utils/tolerance.stanza
m-hilgendorf: Done

utils/design-vars.stanza
```diff
@@ -1,14 +1,17 @@
...
-;==============================================================================
-;========================= Global Checks Parameters ===========================
-;==============================================================================
-public var OPERATING-TEMPERATURE = [0.0 25.0] ; the [min, max] operating temperature of the design, in Celsius.
+public pcb-enum ocdb/design-vars/DensityLevel:
+  DensityLevelA
+  DensityLevelB
+  DensityLevelC
+
+public var OPERATING-TEMPERATURE = [0.0 25.0]
```
bpellenard, line 14: Should we mention the unit?

utils/land-patterns.stanza
```diff
@@ -18,6 +20,55 @@ defn ensure-even-positive! (value: Int, name: String):
   if value < 2 or value % 2 == 1:
     fatal("%_ only supports a positive even number of pins." % [name])
 
+; ====== Silk screen helpers ===================================================
+; Create a polarity marker relative to a pad.
+; pda
```
bpellenard, line 25: pda?

utils/design-vars.stanza
m-hilgendorf: I think we assume C

utils/land-patterns.stanza
m-hilgendorf: Good catch, fixed.

utils/land-patterns.stanza
```diff
@@ -18,6 +20,55 @@ defn ensure-even-positive! (value: Int, name: String):
...
+        x(pad-dims) * -0.5 + x(center(pad-loc)) - SOLDER-MASK-REGISTRATION
+        y(pad-dims) *  0.5 + y(center(pad-loc)) + R
+      )
+    
+    val [p1, p2] = [pad-pose(), part-pose()] 
+    draw-line $
+      if distance(p1) > distance(p2):
+        p1
+      else: 
+        p2
```
bpellenard, line 69: max(distance, [p1 p2])?

m-hilgendorf: Which usage of `max` is this? I'm not familiar

bpellenard: or maybe `maximum`

utils/tolerance.stanza
```diff
@@ -0,0 +1,60 @@
...
+  import math
+
+; A generic value with some tolerance. May be offcenter (upper tolerance != lower tolerance)
+public pcb-struct ocdb/tolerance/Toleranced:
+  typ: Double,
+  tol+:Double,
+  tol-:Double,
+
+defmethod print (o:OutputStream, t:Toleranced):
+  println(o, "Toleranced(%_ <= %_ <= %_)" % [min-value(t), typ(t), max-value(t)])
```
bpellenard, line 14: `print`

utils/tolerance.stanza
```diff
@@ -0,0 +1,60 @@
+#use-added-syntax(jitx)
+defpackage ocdb/tolerance:
+  import core
+  import collections
+  import math
+
+; A generic value with some tolerance. May be offcenter (upper tolerance != lower tolerance)
+public pcb-struct ocdb/tolerance/Toleranced:
+  typ: Double,
```
bpellenard, line 9: could it be 0.0?

utils/tolerance.stanza
```diff
@@ -0,0 +1,60 @@
...
+public defn tol (typ:Double, tol+:Double, tol-:Double) -> Toleranced:
+  Toleranced(typ, tol+, tol-)
+
+; Create a tolerance with the same upper and lower tolerances
+public defn tol (typ:Double, tol+-:Double) -> Toleranced:
+  tol(typ, tol+-, tol+-)
+
+; Return the upper tolerance, as a percentage
+public defn tol+% (t:Toleranced) -> Double:
+  tol+(t) / typ(t)
```
bpellenard, line 36: careful if `typ(t)` equals `0.0`

utils/tolerance.stanza
```diff
@@ -0,0 +1,60 @@
...
+public defn max-value (t:Toleranced) -> Double:
+  typ(t) + tol-(t)
+
+; Return the min value
+public defn min-value (t:Toleranced) -> Double:
+  typ(t) - tol-(t)
+
+; Return the center of the value range
+public defn center (t:Toleranced) -> Double:
+  0.5 * (min-value(t) + max-value(t))
```
bpellenard, line 52: careful with overflow: `min-value(t) + 0.5 * (max-value(t) - min-value(t))`

utils/tolerance.stanza
```diff
@@ -0,0 +1,60 @@
...
+
+; Check if a value is within the tolerance's range.
+public defn in-range? (t:Toleranced, value:Double) -> True|False:
+  value >= min-value(t) and value <= max-value(t)
+ 
+ ; Return the difference between max and min of the toleranced values
```
bpellenard, line 58: Returns?

utils/tolerance.stanza
```diff
@@ -0,0 +1,60 @@
+#use-added-syntax(jitx)
+defpackage ocdb/tolerance:
```
bpellenard, line 2: should we create unit tests for the packages in this repo?

utils/tolerance.stanza
m-hilgendorf: It's very unlikely. But we also can't use `with: (ensure)` in pcb-structs.

utils/tolerance.stanza
m-hilgendorf: Even with doubles?

utils/tolerance.stanza
m-hilgendorf: I think if `typ == 0.0` then this is still valid, the result would be NaN

utils/tolerance.stanza
m-hilgendorf: done

utils/tolerance.stanza
m-hilgendorf: done`

utils/tolerance.stanza
m-hilgendorf: done

utils/tolerance.stanza
m-hilgendorf: Done - also good call, it found a few bugs

utils/tolerance.stanza
```diff
@@ -0,0 +1,65 @@
...
+; Create a tolerance with the same upper and lower tolerances
+public defn tol (typ:Double, tol+-:Double) -> Toleranced:
+  tol(typ, tol+-, tol+-)
+
+; Returns the upper tolerance, as a percentage
+public defn tol+% (t:Toleranced) -> Double:
+  tol+(t) / typ(t)
+
+; Returns the lower tolerance, as a percentage
+public defn tol-% (t:Toleranced) -> Double:
```
bpellenard, line 44: add unit tests for this function

utils/tolerance.stanza
bpellenard: ok, I would like to see this edge case covered in the unit tests

utils/tolerance.stanza
```diff
@@ -0,0 +1,65 @@
...
+; Returns the center of the value range
+public defn center (t:Toleranced) -> Double:
+  min-value(t) + 0.5 * (max-value(t) - min-value(t))
+
+; Check if a value is within the tolerance's range.
+public defn in-range? (t:Toleranced, value:Double) -> True|False:
+  value >= min-value(t) and value <= max-value(t)
+ 
+ ; Returns the difference between max and min of the toleranced values
+public defn tolerance-range (t:Toleranced) -> Double:
```
bpellenard, line 64: add unit test

utils/tolerance.stanza
```diff
@@ -0,0 +1,65 @@
...
+  import collections
+  import math
+
+; A generic value with some tolerance. May be offcenter (upper tolerance != lower tolerance)
+public pcb-struct ocdb/tolerance/Toleranced:
+  typ: Double,
+  tol+:Double,
+  tol-:Double,
+
+defmethod print (o:OutputStream, t:Toleranced):
```
bpellenard, line 13: add in unit test, not as `print` but as `to-string`, and call `#EXPECT` on the result

m-hilgendorf: done

utils/tolerance.stanza
m-hilgendorf: done

utils/tolerance.stanza
m-hilgendorf: done

bpellenard (approved):
It looks great @m-hilgendorf! Thank you very much for the updates.

utils/tolerance.stanza
CuppoJava: Better to crash with `fatal` than to return NaN.

m-hilgendorf: It actually returns infinity in this case. Should we still crash?

utils/land-patterns.stanza
```diff
@@ -911,22 +966,22 @@ public pcb-landpattern INDC1005X55 :
...
 public pcb-landpattern CAPT2012X125 :
   make-ipc-two-pin-landpattern("0805")
-  layer(Silkscreen("pol", Top)) = PlusSymbol(-2.4, 0.0)
+  draw-polarity-marker(self.p[1])
   ref-label()
 
 public pcb-landpattern CAPT3216X160 :
   make-ipc-two-pin-landpattern("1206")
-  layer(Silkscreen("pol", Top)) = PlusSymbol(-3.0, 0.0)
+  draw-polarity-marker(self.p[1])
```
CuppoJava, line 984: See later comment.

utils/land-patterns.stanza
```diff
@@ -1034,3 +1089,154 @@ public pcb-landpattern smd-testpoint-pkg (d:Double) :
   layer(Courtyard(Top)) = Circle(d / 2.0)
   layer(Silkscreen("f-silk", Top)) = Circle(d / 2.0)
   ref-label()
+
+public defn make-two-pin-chip-landpattern (length:Toleranced,
```
CuppoJava, line 1093: Please add a comment describing what this function is supposed to do. Then we can judge whether or not the API is appropriate. 

For example: Is this generator responsible for drawing the polarity marker when it is necessary? 

Currently it seems the answer is: It depends. 
1) When modeling a non-polarized component, this generator will handle the polarity marker appropriately. (By doing nothing.)
2) When modeling a polarized component, with pins a and c, this generator will handle the polarity marker appropriately. (By drawing it.)
3) When modeling a polarized component, with pins p[0] and p[1], (e.g. CAPT3216X160) then it is the caller's responsibility to pass `false` for the `polarized?` argument, and then manually call `draw-polarity-marker` themselves.

Point 3 when written out, makes it seem like this would be a tricky API to use.

utils/land-patterns.stanza
```diff
@@ -18,6 +20,58 @@ defn ensure-even-positive! (value: Int, name: String):
   if value < 2 or value % 2 == 1:
     fatal("%_ only supports a positive even number of pins." % [name])
 
+; ====== Silk screen helpers ===================================================
+; Create a polarity marker relative to a pad.
+; 
+; pad-1: the pad to mark for the polarity 
+; part-dims?: optional. If provided, the generator will use these to determine where
```
CuppoJava, line 27: This utility seems to assume that the current landing pattern is centered at the origin. Is that true? 

Can we remove this assumption? If not, this assumption makes this utility less applicable as a general user-facing utility.

utils/land-patterns.stanza
```diff
@@ -18,6 +20,58 @@ defn ensure-even-positive! (value: Int, name: String):
...
+;             to place the pad, and guarantee it is not underneath the part.
+public defn draw-polarity-marker (pad-1:JITXObject, part-dims?:Dims|False) :
```
CuppoJava, line 29: Can we have a screen shot of what this function is supposed to generate?

utils/land-patterns.stanza
```diff
@@ -18,6 +20,58 @@ defn ensure-even-positive! (value: Int, name: String):
...
+    defn draw-line (origin:Pose):
+      val line-width  = min-silk-width
+      val line-length = line-width * 3.0
+      val line = Line(line-width, [
+        Point(min-silk-width * 0.5, 0.0)
+        Point(line-width + min-silk-width * 0.5, 0.0)  
+      ])
+      layer(Silkscreen("pol", Top)) = origin * line
+
+    defn part-pose () :
```
CuppoJava, line 43: I'm not sure what the `part-dims?` argument is controlling here, can you describe it?

CuppoJava (changes requested):
Clarification needed on some APIs.

utils/land-patterns.stanza
CuppoJava: Thanks for point out that these are actually calling: make-ipc-two-pin-landpattern instead of the newer make-two-pin-chip-landpattern.

utils/tolerance.stanza
CuppoJava: Yes. Crashing is good. I cannot imagine a scenario where a user will rely upon the +Inf for his code to work.

m-hilgendorf: Fixed, these functions now call `fatal`.

utils/land-patterns.stanza
```diff
@@ -1178,7 +1180,7 @@ public defn make-two-pin-chip-landpattern (length:Toleranced,
   make-two-pin-chip-landpattern(
     length, width, lead-length, lead-width?, 
     FABRICATION-TOLERANCE, PLACEMENT-TOLERANCE, 
-    density-level, polarized?,
+    density-level, polarized?, polarized?
```
bpellenard, line 1183: are you going to use, in the future, `make-two-pin-chip-landpattern` with different values for `anode-and-cathode?` and `polarized?`?

m-hilgendorf: Yes.

m-hilgendorf:
Updated screen shot showing polarity marker shape and distance:

Screenshot from 2021-07-26 12-26-56

CuppoJava (approved):
Looks great. Thank you!

PR #89, opened 2021-07-27
Feature/jitx 522 add missing pins 1

RajeshJITX:
Added NSS pin to SPI in bundles.stanza.
**The above statement is incorrect. Added NSS pin to UART and USART in bundles.stanza**
Rewrote the logic to print SPI and UART
Bug fix for printing I2C. In some cases we were picking up SMBA pin which is supposed to be ignored
Changed fatal to Exception so the caller can decide what to do if a file parsing fails.

### DIFF
--- a/utils/bundles.stanza
+++ b/utils/bundles.stanza
@@ -88,7 +88,7 @@ public pcb-enum ocdb/bundles/UARTPins :
   UART-RTS
   UART-CK
   UART-DE
-  UART-NSS
+  UART-CS
 
 public pcb-bundle uart (pins:Tuple<UARTPins>) :
   pin rx
@@ -103,7 +103,7 @@ public pcb-bundle uart (pins:Tuple<UARTPins>) :
       UART-RTS : make-pin(`rts)
       UART-CK : make-pin(`ck)
       UART-DE : make-pin(`de)
-      UART-NSS : make-pin(`nss)
+      UART-CS : make-pin(`cs)
 
 public defn uart ():
   uart([])
@@ -121,7 +121,7 @@ public pcb-bundle usart (pins:Tuple<ocdb/bundles/UARTPins>) :
       UART-RTS : make-pin(`rts)
       UART-CK : make-pin(`ck)
       UART-DE : make-pin(`de)
-      UART-NSS : make-pin(`nss)
+      UART-CS : make-pin(`cs)
 
 public defn usart ():
   usart([])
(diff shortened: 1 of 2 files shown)


utils/bundles.stanza
```diff
@@ -140,6 +143,7 @@ public pcb-bundle spi  (pins:Tuple<SPIPins>):
       SPI-CS  : make-pin(`cs)
       SPI-CIPO  : make-pin(`cipo)
       SPI-COPI  : make-pin(`copi)
+      SPI-SDIO  : make-pin(`sdio)
```
d-haldane, line 146: Where does this pin come from? Is it in standard SPI?

m-hilgendorf (approved):
Looks good!

utils/bundles.stanza
RajeshJITX: It looks like a standard SPI. It is present in many files and it shows up as 
```
STM32WLE5CBUx.xml:		<Signal Name="SPI1_NSS"/>
STM32WLE5CBUx.xml:		<Signal Name="SPI2_NSS"/>
STM32MP151CAAx.xml:		<Signal Name="SPI5_NSS"/>
```

d-haldane: I was asking about `SDIO` . NSS for SPI you already capture and map to spi.cs, correct?

RajeshJITX: My description in PR is incorrect. I added a new NSS pin for UART and USART. NSS in SPI maps to cs and that has not changed

d-haldane: OK. NSS should show up in JITX as CS, so UART-CS and USART-CS should be the pins in bundles.stanza.

Still curious about SDIO in SPI as a separate issue.

RajeshJITX: First about NSS. I will rename the pin. The current pin names in UART/USART are
```
  UART-DTR
  UART-CTS
  UART-DCD
  UART-RI
  UART-DST
  UART-RTS
  UART-CK
  UART-DE
  UART-NSS
```

I will change UART-NSS to UART-CS. Any other changes? Do any of the other pins also map to UART-CS?

Second about SDIO. I have not seen it appear anywhere in STM files. The only things show up are sck, sdo, sdi and sdo. Please let me know if we should change anything else

electricfungi (approved):
Approved understanding it's a work in progress towards final bundles.stanza

PR #92, opened 2021-07-30
Call set-default-rules where appropriate

m-hilgendorf:


utils/defaults.stanza
```diff
@@ -63,25 +63,24 @@ public pcb-board default-board (num-layers:Int, outline:Shape) :
 
 ; Creates a design using the default board. 
 public defn make-default-board (module:Instantiable, num-layers:Int, outline:Shape) :
-  
+  set-rules(ocdb/rules/bay-area-circuits-conservative-rules)
```
PhilippeFerreiraDeSousa, line 66: Shouldn't we remove this from here all together if we need to call it in the design file anyway? Or is it just for rectangular boards and some other shapes that we need it in the design file?

PhilippeFerreiraDeSousa: Other designs than tutorial and mcu work thanks to this call to set-rules in here?

m-hilgendorf: I don't know actually. I think its hazardous because it silently overwrites rule in subsequent calls. So maybe we should remove it.

PhilippeFerreiraDeSousa (approved):
We can merge this fix now. And make a ticket / plan a meeting for later steps?

m-hilgendorf:
I'm closing because this has some subtleties and we have a different workaround.

PR #93, opened 2021-07-30
Hardcode line width for LineRectangle

m-hilgendorf:


PhilippeFerreiraDeSousa (approved):
Okay, if you think it is best for now

PR #94, opened 2021-08-01
Feature/jitx 522 cleanup pin names

RajeshJITX:
Added logic consistent with python script to ensure that the pin names are identical


PhilippeFerreiraDeSousa:
It could be useful to *document* the behavior with a test and the edge case values to see what was chosen for their stanza names :
```
  import lang-utils  ; for kvs

deftest test-extract-pin-name :
  val io = to-hashtable<String, String> $ [
    "input1" => "output1",
    "input2" => "output2,
    ...
  ]
  for [input, output] in kvs(ios) do :
    #ASSERT(extract-pin-name[input] == output)
```

RajeshJITX:
I have these tests. I will add them

PhilippeFerreiraDeSousa (approved):
Thank you for the behaviour documentation with the tests!

tests/test-stm-pin-parsing.stanza
```diff
@@ -0,0 +1,481 @@
...
+    "VSS_PLL2"                     => "VSS_PLL2",
+    "VSS_USBHS"                    => "VSS_USBHS"
+  ]
+
+  for [input, output] in kvs(pinnames) do :
+    var ret = extract-pin-name(input)
+    ret = to-string(ref-ify(ret))
+    #ASSERT(ret == output)
+
+defn kvs<?K,?V> (xs:Seqable<KeyValue<?K,?V>>) :
```
PhilippeFerreiraDeSousa, line 480: I copied pasted `kvs` from lang-utils because I had isolated the test as pure stanza code. Not relevant now that we have jitx as dependency. It doesn't matter much that is just 2 lines

tests/test-stm-pin-parsing.stanza
```diff
@@ -0,0 +1,481 @@
...
+    "PG6"                          => "PG[6]",
+    "PG7"                          => "PG[7]",
+    "PG8"                          => "PG[8]",
+    "PG9"                          => "PG[9]",
+    "PH0 - OSC_IN"                 => "PH[0]",
+    "PH0"                          => "PH[0]",
+    "PH0-OSC_IN (PH0)"             => "PH[0]",
+    "PH0-OSC_IN"                   => "PH[0]",
+    "PH0/OSC_IN"                   => "PH[0]",
+    "PH1 - OSC_OUT"                => "PH[1]",
```
bpellenard, line 308: Is it the expected result? `"PH1 - OSC_OUT"=> "PH[1]"`

tests/test-stm-pin-parsing.stanza
```diff
@@ -0,0 +1,481 @@
...
+    "VSSA/VREF-"                   => "VSSA",
+    "VSSDSI"                       => "VSSDSI",
+    "VSSRF"                        => "VSSRF",
+    "VSSSD"                        => "VSSSD",
+    "VSSSD/VREFSD-"                => "VSSSD",
+    "VSSSMPS"                      => "VSSSMPS",
+    "VSS_ANA"                      => "VSS_ANA",
+    "VSS_DSI"                      => "VSS_DSI",
+    "VSS_PLL"                      => "VSS_PLL",
+    "VSS_PLL2"                     => "VSS_PLL2",
```
bpellenard, line 471: `"VCAP1"=> "VCAP[1]"` before.
why not `"VSS_PLL2" => "VSS_PLL[2]"` here?

scripts/cubemx-importer-utils.stanza
```diff
@@ -0,0 +1,152 @@
...
+defn sanitize-pin-name (s:String) -> String:
+  var s*: String = s
+  val chars = " (/-"
+  for c in chars do:
+    match(index-of-char(s*, c)):
+      (i:Int): s* = s*[0 to i]
+      (_:?): false
+  s*
+
+defn get-numbers-in-string (s:String) -> Vector<[Int, Int]>:
```
bpellenard, line 17: I would also suggest to have unit tests for this function

tests/test-stm-pin-parsing.stanza
RajeshJITX: None of the voltages have indexes associated with them

scripts/cubemx-importer-utils.stanza
```diff
@@ -0,0 +1,152 @@
...
+    ret
+  else if ( prefix?(ret1, "UCPD1_DBCC")):
+    ; These have array length of 1
+    val [start, end] = arr[1]
+    val num = ret1[start to end]
+    val ret = to-string("UCPD1_DBCC[%_]" % [num])
+    ;println("2: %_ ==> %_" % [raw, ret])
+    ret
+  else if ( length(arr) == 0 or
+            prefix?(ret1, "VDD") or
```
bpellenard, line 66: I would suggest to extract voltage prefixed in a container with an explicit name: `voltage-...`, and use this container here

tests/test-stm-pin-parsing.stanza
RajeshJITX: Yes. It is. They all entered by humans and they all map to PH[0]

PR #97, opened 2021-08-02
removed SDI and SDO from SPI bundle, as these are duplicates of CIPO and COPI.

electricfungi:


RajeshJITX:
These changes could break some generated as well existing code. If you want, you can create a temp branch like temp/suggested-changes and I will pick up changes from these branches

d-haldane:
For these names, SDI maps to MISO, and SDO maps to MOSI.

d-haldane:
Or maybe not. Here's the source: https://www.oshwa.org/a-resolution-to-redefine-spi-signal-names We need a chart to figure out what to call what.

electricfungi:
Closing so we match with https://www.oshwa.org/a-resolution-to-redefine-spi-signal-names

PR #98, opened 2021-08-05
IPC Compliant Quad Landpatterns

m-hilgendorf:
(blocked on internal support for `expand(...)`

Examples (generated using the code in tests/test-land-patterns.stanza):

qfn-final

qfp-final

The following APIs are introduced:

```
public defn make-quad-landpattern (num-pins-west: Int,  ; the number of pins on the west face
                                   num-pins-south: Int, ; the number of pins on the south face
                                   num-pins-east: Int,  ; the number of pins on the east face
                                   num-pins-north: Int, ; the number of pins on the north face
                                   pitch-west:Double,   ; the pithc on the west face
                                   pitch-south:Double,  ; the pitch on the south face
                                   pitch-east:Double,   ; the pitch on the east face
                                   pitch-north:Double,  ; the pitch on the north face
                                   component-length: Toleranced, ; the distance from pad-edge to pad-edge on the exterior of the part in the x direction
                                   component-width: Toleranced, ; the distance from pad-edge to pad-edge on the exterior of the part in the y direction
                                   package-length: Toleranced,  ; the size of the package in the x direction
                                   package-width: Toleranced,  ; the size of hte package in the y direction
                                   lead-length: Toleranced, ; the dimension of the leads in the x direction on the east/west faces
                                   lead-width: Toleranced, ; the dimension of the leads in the y direction on the east/west faces
                                   lead-type:  LandProtrusionType, ; the type of the leads
                                   density-level: DensityLevel   ; the density level of the design
                                   ):
```

This is the base generator for quad land patterns (pads on all four edges). 

Some additional helpers to simplify calling this generator:

```
public defn make-quad-landpattern (num-pins:Int, 
                                   pitch:Double, 
                                   outer-xy:Toleranced, 
                                   package-xy:Toleranced, 
                                   lead-x:Toleranced, 
                                   lead-y:Toleranced, 
                                   lead-type:LandProtrusionType):

public defn make-qfn-landpattern (num-pins:Int, 
                                  pitch:Double, 
                                  package-size:Toleranced, 
                                  lead-length:Toleranced, 
                                  lead-width:Toleranced):

public defn make-qfn-landpattern (num-pins:[Int, Int, Int, Int], 
                                  pitch:Double, 
                                  package-size:Toleranced, 
                                  lead-length:Toleranced, 
                                  lead-width:Toleranced):
           
public defn make-qfp-landpattern (num-pins:Int, 
                                  pitch:Double, 
                                  component-size:Toleranced, 
                                  package-size:Toleranced, 
                                  lead-length:Toleranced, 
                                  lead-width:Toleranced):

```

### DIFF
--- a/tests/test-land-patterns.stanza
+++ b/tests/test-land-patterns.stanza
@@ -122,4 +122,4 @@ pcb-module main-module:
 
 set-rules(sierra-adv-rules)
 make-default-board(main-module, 2, Rectangle(12.0, 12.0))
-view(QFP)
+view-board()
--- a/tests/test-land-patterns.stanza
+++ b/tests/test-land-patterns.stanza
@@ -31,12 +31,12 @@ pcb-component QFP:
 
 pcb-component QFN:
   pcb-landpattern MLPQ-UT8:
-    make-qfn-landpattern([1, 1, 3, 3],    ; number of pins on each side (Up, Down, Left, Right)
+    make-qfn-landpattern([3, 1, 3, 1],    ; number of pins on each side (Up, Down, Left, Right)
                          0.4,             ; pitch
                          tol(1.5, 0.0),   ; package dimensions 
                          tol(0.35, 0.05), ; lead size, x direction (for p[1])
                          tol(0.20, 0.05)) ; lead size, y direction (for p[1])
-    
+
   port p: pin[1 through 8]
   landpattern = MLPQ-UT8(
     for n in 1 through 8 do:
@@ -122,4 +122,4 @@ pcb-module main-module:
 
 set-rules(sierra-adv-rules)
 make-default-board(main-module, 2, Rectangle(12.0, 12.0))
-view-board()
+view(QFP)
--- a/tests/test-land-patterns.stanza
+++ b/tests/test-land-patterns.stanza
@@ -12,6 +12,53 @@ defpackage ocdb/test/land-patterns:
   import ocdb/rules
   import ocdb/generic-components
   import ocdb/symbols
+  import ocdb/land-protrusions
+
+pcb-component QFP:
+  pcb-landpattern LQFP48:
+    make-quad-landpattern(
+      24, 24   ; pins on north/south and east/west faces
+      0.5, 0.5 ; pitch on north/south and east/west faces
+      min-typ-max(8.8, 9.0, 9.2), min-typ-max(8.8, 9.0, 9.2) ; outer-x, outer-y ("D" and "E")
+      min-typ-max(6.8, 7.0, 7.2), min-typ-max(6.8, 7.0, 7.2) ; part-x, part-y (D1, E1)
+      min-typ-max(0.45, 0.6, 0.75), min-typ-max(0.17, 0.22, 0.27) ; lead-x, lead-y (L, b)
+      SmallGullWingLeads
+    )
+  port p : pin[[1]]
+  val lp = LQFP48
+  landpattern = 
+    lp(p[1] => lp.p[1])
+  make-box-symbol()
+
+pcb-component QFN:
+  pcb-landpattern MLPQ-UT8:
+    make-quad-landpattern(
+      2,  ; pins on north/south faces
+      6,  ; pins on east/west faces
+      0.4 ; pitch on north-south faces
+      0.4 ; pitch on east-west faces
+      tol(1.5, 0.0) ; outer-x
+      tol(1.5, 0.0) ; outer-y
+      tol(1.5, 0.0) ; part-x
+      tol(1.5, 0.0) ; part-y
+      min-typ-max(0.30, 0.35, 0.40) ; lead-dim-1
+      min-typ-max(0.15, 0.20, 0.25) ; lead-dim-2
+      QuadFlatNoLeads
+    )
+  val lp = MLPQ-UT8
+  port p: pin[[1 through 8]]
+  landpattern = 
+    lp(
+      p[1] => lp.p[1]
+      p[2] => lp.p[2]
+      p[3] => lp.p[3]
+      p[4] => lp.p[4]
+      p[5] => lp.p[5]
+      p[6] => lp.p[6]
+      p[7] => lp.p[7]
+      p[8] => lp.p[8]
+    )
+  make-box-symbol()
 
 pcb-component SOIC-CMP:
   pcb-landpattern SO8N:
@@ -81,10 +128,14 @@ pcb-module main-module:
   inst e1: example-1
   inst e2: SOIC-CMP
   inst e3: example-2
+  inst e4: QFN
+  inst e5: QFP
   place(e1) on Top
   place(e2) on Top
   place(e3) on Top 
+  place(e4) on Top
+
 set-rules(sierra-adv-rules)
 make-default-board(main-module, 2, Rectangle(12.0, 12.0))
 ; view-board()
-view(SOIC-CMP)
+view(QFP)
--- a/tests/test-land-patterns.stanza
+++ b/tests/test-land-patterns.stanza
@@ -24,10 +24,12 @@ pcb-component QFP:
       min-typ-max(0.45, 0.6, 0.75), min-typ-max(0.17, 0.22, 0.27) ; lead-x, lead-y (L, b)
       SmallGullWingLeads
     )
-  port p : pin[[1]]
+  port p: pin[1 through 48]
   val lp = LQFP48
-  landpattern = 
-    lp(p[1] => lp.p[1])
+  landpattern = lp(
+    for n in 1 through 48 do:
+      p[n] => lp.p[n]
+  )
   make-box-symbol()
 
 pcb-component QFN:
--- a/tests/test-land-patterns.stanza
+++ b/tests/test-land-patterns.stanza
@@ -16,50 +16,32 @@ defpackage ocdb/test/land-patterns:
 
 pcb-component QFP:
   pcb-landpattern LQFP48:
-    make-quad-landpattern(
-      24, 24   ; pins on north/south and east/west faces
-      0.5, 0.5 ; pitch on north/south and east/west faces
-      min-typ-max(8.8, 9.0, 9.2), min-typ-max(8.8, 9.0, 9.2) ; outer-x, outer-y ("D" and "E")
-      min-typ-max(6.8, 7.0, 7.2), min-typ-max(6.8, 7.0, 7.2) ; part-x, part-y (D1, E1)
-      min-typ-max(0.45, 0.6, 0.75), min-typ-max(0.17, 0.22, 0.27) ; lead-x, lead-y (L, b)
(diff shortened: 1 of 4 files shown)


utils/land-patterns.stanza
```diff
@@ -1071,7 +1091,154 @@ defn ipc-formula (L:Toleranced, ; the distance from edge-of-lead to edge-of-lead
...
+  make-quad-landpattern(num-pins / 4, num-pins / 4, num-pins / 4, num-pins / 4, 
+                        pitch, pitch, pitch, pitch, 
+                        component-size, component-size, 
+                        package-size, package-size, 
+                        lead-length, lead-width, lead-type, DensityLevelC)
+  
+public defn make-qfn-landpattern (num-pins:Int, 
+                                  pitch:Double, 
+                                  package-size:Toleranced, 
+                                  lead-length:Toleranced,
```
d-haldane, line 1210: 'lead' is the wrong terminology for QFNs, which are 'leadless'. You used the JEDEC spec for the soic, use it again here. Maybe the right term is `package-pin-length` ?

utils/land-patterns.stanza
```diff
@@ -1071,7 +1091,154 @@ defn ipc-formula (L:Toleranced, ; the distance from edge-of-lead to edge-of-lead
...
+    ])
+
+    ;----------------------------------------------------------
+    ; Place the reference label
+    val text-y = 0.5 * (y(cy-sz) + MIN-SILKSCREEN-TEXT-HEIGHT)
+    ref-label(0.0, text-y)
+
+public defn make-quad-landpattern (num-pins:Int, 
+                                   pitch:Double, 
+                                   component-size:Toleranced,
```
d-haldane, line 1193: What does component size mean? Not a standard term I'm familiar with. Is it leadspan for QFPs?

utils/land-patterns.stanza
```diff
@@ -1071,7 +1091,154 @@ defn ipc-formula (L:Toleranced, ; the distance from edge-of-lead to edge-of-lead
...
+public defn make-qfn-landpattern (num-pins:[Int, Int, Int, Int], 
+                                  pitch:Double, 
+                                  package-size:Toleranced, 
+                                  lead-length:Toleranced, 
+                                  lead-width:Toleranced):
+  make-quad-landpattern(num-pins[0], num-pins[1], num-pins[2], num-pins[3], 
+                        pitch, pitch, pitch, pitch,
+                        package-size, package-size,
+                        package-size, package-size,
+                        lead-length, lead-width, QuadFlatNoLeads, DensityLevelC)
```
d-haldane, line 1227: This is incorrect. The default should pull the global parameter for density level, not hardcode it to DensityLevelC

utils/land-patterns.stanza
m-hilgendorf: Good catch, this got hit by a systematic rename.

utils/land-patterns.stanza
m-hilgendorf: This is repeated from the SO generator. What is "leadspan?"

utils/land-patterns.stanza
m-hilgendorf: Thanks!

utils/land-patterns.stanza
d-haldane: Then SO generator needs to be updated. I think `component-size` means leadspan, but use this figure to double check: https://www.pcb-3d.com/models/qfp80p900x900x160-32n_jedec_ms-026bba/

utils/land-patterns.stanza
m-hilgendorf: is `terminal-length` alright?

d-haldane: Yeah, sounds fine.

utils/land-patterns.stanza
m-hilgendorf: done

utils/land-patterns.stanza
m-hilgendorf: fixed

tests/test-land-patterns.stanza
```diff
@@ -12,6 +12,37 @@ defpackage ocdb/test/land-patterns:
...
+    make-qfn-landpattern([3, 1, 3, 1],    ; number of pins on each side (Up, Down, Left, Right)
+                         0.4,             ; pitch
+                         tol(1.5, 0.0),   ; package dimensions 
+                         tol(0.35, 0.05), ; lead size, x direction (for p[1])
+                         tol(0.20, 0.05)) ; lead size, y direction (for p[1])
+
+  port p: pin[1 through 8]
+  landpattern = MLPQ-UT8(
+    for n in 1 through 8 do:
+      p[n] => MLPQ-UT8.p[n]
```
bpellenard, line 43: I am curious about if `map({p[_0] => MLPQ-UT8.p[_0]}, 1 through 8)` is valid

utils/land-patterns.stanza
```diff
@@ -1306,27 +1308,27 @@ public defn make-n-pin-soic-landpattern (num-pins:Int,                 ; number
...
                               terminal-width,
-                              DensityLevelC)
+                              DENSITY-LEVEL)
 
 public defn make-two-pin-chip-landpattern (length:Toleranced, ; the length (vertical) dimension of the package
                                            width:Toleranced,  ; the width (horizontal) dimension of the package
                                            lead-length:Toleranced, ; the length (vertical) dimension of the conductors/leads of the package
                                            lead-width?:False|Toleranced, ; the width (horizontal) dimension of the conductors/leads of the package, if different from `length`
-                                           density-level:DensityLevel,   ; the density level of the design (DensityLevelA|DensityLevelB|DensityLevelC)
+                                           density-level:DensityLevel,   ; the density level of the design (DensityLevelA|DensityLevelB|DENSITY-LEVEL)
```
d-haldane, line 1331: Looks like find/replace overstepped here.

utils/land-patterns.stanza
```diff
@@ -1102,8 +1102,8 @@ public defn make-quad-landpattern (num-pins-west: Int,  ; the number of pins on
                                    pitch-south:Double,  ; the pitch on the south face
                                    pitch-east:Double,   ; the pitch on the east face
                                    pitch-north:Double,  ; the pitch on the north face
-                                   component-length: Toleranced, ; the distance from pad-edge to pad-edge on the exterior of the part in the x direction
-                                   component-width: Toleranced, ; the distance from pad-edge to pad-edge on the exterior of the part in the y direction
+                                   lead-span-1: Toleranced, ; the distance from pad-edge to pad-edge on the exterior of the part in the x direction
```
d-haldane, line 1105: why `lead-span-1` instead of `lead-span-x`?

m-hilgendorf: Was basing it off the doc shared in slack. We can use `x` instead.

tests/test-land-patterns.stanza
m-hilgendorf: I think `map` is lazy so that wouldn't work necessarily

utils/land-patterns.stanza
m-hilgendorf: thanks, fixed

utils/land-patterns.stanza
```diff
@@ -1071,7 +1091,156 @@ defn ipc-formula (L:Toleranced, ; the distance from edge-of-lead to edge-of-lead
...
+public defn make-quad-landpattern (num-pins:Int, 
+                                   pitch:Double, 
+                                   lead-span:Toleranced, 
+                                   package-size:Toleranced, 
+                                   lead-length:Toleranced, 
+                                   lead-width:Toleranced, 
+                                   lead-type:LandProtrusionType):
+  ensure-divisible!(num-pins, 4, "make-quad-landpattern(num-pins, ...)")
+  if pitch <= 0.0:  
+    fatal("make-quad-landpattern(...) failed: pitch must be greater than 0. Called with: %_." % [pitch])
```
bpellenard, line 1200: should not it be in the lowest level function?

m-hilgendorf: Good point, I'll update the check.

utils/land-patterns.stanza
```diff
@@ -1071,7 +1091,156 @@ defn ipc-formula (L:Toleranced, ; the distance from edge-of-lead to edge-of-lead
...
+                                   lead-span-1: Toleranced, ; the distance from pad-edge to pad-edge on the exterior of the part in the x direction
+                                   lead-span-2: Toleranced, ; the distance from pad-edge to pad-edge on the exterior of the part in the y direction
+                                   package-length: Toleranced,  ; the size of the package in the x direction
+                                   package-width: Toleranced,  ; the size of hte package in the y direction
+                                   lead-length: Toleranced, ; the dimension of the leads in the x direction on the east/west faces
+                                   lead-width: Toleranced, ; the dimension of the leads in the y direction on the east/west faces
+                                   lead-type:  LandProtrusionType, ; the type of the leads
+                                   density-level: DensityLevel   ; the density level of the design
+                                   ):
+  val lead-fillets = lead-fillets(lead-type, density-level)
```
bpellenard, line 1118: `fatal("make-quad-landpattern(...) failed: pitch-west must be greater than 0. Called with: %_." % [pitch-west]) when pitch-west <= 0.0`

m-hilgendorf: I actually expect this to be less useful as an error message since the common case is all pitches being shared

bpellenard: why 4 parameters if they are shared?

m-hilgendorf: They may be different, but I don't expect that to be the common case.

tests/test-land-patterns.stanza
ariel-hi: Bertrand is a quick learner with `_0`!
Here you can use `do` instead of `map`. The key-value pairs aren't actually returned, they are macros to ESIR calls.

ariel-hi: I would keep it as is -- the way Mike wrote it follows our ESIR design standards since curlies aren't beginner-friendly.

d-haldane (approved):

utils/land-patterns.stanza
```diff
@@ -97,6 +103,17 @@ defn chamfered (anchor:Anchor, width:Double, height:Double, r:Double):
 ;==============================================================================
 ;=========================== Geometric Helpers ================================
 ;==============================================================================
+defn shift-locs (x:Double, y:Double, s:Seqable<Pose>):
+  seq({loc(x, y) * _}, s)
+
+defn max-distance (p1:Point, p2:Point) -> Point:
+  defn radius (p:Point):
```
bpellenard, line 110: defn radial-coordinate (p:Point) -> Double :

bpellenard (approved):
Looks good to me

PR #109, opened 2021-08-30
Remove extra ':' in CC2640/module definition

m-hilgendorf:


PhilippeFerreiraDeSousa (approved):
You see, that would happen a lot without macro evaluation tests

m-hilgendorf:
Thanks! I'm not sure why this was skipped originally.

components/texas-instruments/CC2640.stanza
```diff
@@ -121,11 +121,10 @@ public pcb-component component :
   supports high-drive :
     high-drive.high-drive => self.DIO_[17]
 
-;<test>skip<test>
```
PhilippeFerreiraDeSousa, line 124: Actually I think that macros of modules are not evaluated because those are bigger, longer to run and they are mostly just instantiating the component defined above them so that would duplicate the macro evaluation of components.

So I don't think that this test skip instruction was doing anything

PR #110, opened 2021-08-30
Integration tests using docker image pkgs instead of ocdb repo source files

PhilippeFerreiraDeSousa:
Integration tests were using jitx-client docker pkgs instead of the source files from the ocdb repo of the PR

### DIFF
--- a/.github/jitx-client/scripts/run-ocdb-tests.sh
+++ b/.github/jitx-client/scripts/run-ocdb-tests.sh
@@ -9,12 +9,24 @@ jitx check-install
 echo "/root/.jitx/user.params:"
 cat /root/.jitx/user.params
 
+#==============================================
+#==== public pcb-* macro evaluation tests =====
+#==============================================
+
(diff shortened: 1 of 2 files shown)


.github/jitx-client/scripts/run-ocdb-tests.sh
```diff
@@ -9,13 +9,25 @@ jitx check-install
...
 echo "Launching pcb object tests..."
 jitx run-test test/evaluate/api
 cd ..
 
+#==============================================
+#====== Unit-tests and integration-tests ======
+#==============================================
+
+# Those tests need to take into account the root stanza.proj so that they find the source files from the ocdb repo and not a pkg from the jitx-client docker image
+export STANZA_CONFIG=/app/open-components-database/
```
m-hilgendorf, line 29: Is there a .stanza file here? You can pass `stanza.proj` files as args to `jitx` as well.

PhilippeFerreiraDeSousa: Oh no there is none, this is wrong

m-hilgendorf (approved):

PR #111, opened 2021-09-02
Temp/ah/mcu json

ariel-hi:
JSON:
Renamed top-level field from `support` to `supports`
Removed `supports-options?` field

Stanza:
Renamed STM Stanza fields
Added Stanza printers

JITX:
Added Stanza to JITX conversions (ocdb bundles largely commented out)


utils/stm-to-jitx.stanza
```diff
@@ -0,0 +1,373 @@
...
+    defn generate-mapped-ref (mapping:STMSupportMapping) -> JITXObject :
+      inside pcb-module :
+        require mygpio:gpio
+        match(require(mapping)) :
+          (req:String) :
+            val req-bundle = bundle-table[req]
+            require p0 : req-bundle
+            dot(p0, ref-ify(pin(mapping)))
+          (req:False) :
+            dot(self, ref-ify(pin(mapping)))
```
PhilippeFerreiraDeSousa, line 99: The stanza code that does database to esir component mapping should be as general as possible to be reuse for other types of components with supports. json -> supports statements can be general.

`ref-ify` was a function to format refs as they appeared on the xml we scraped the data from for stms, we should use the default print of a ref to to represent a ref in json like "a.b[2].c.d[3]". I wrote a function to parse this back into a Ref

```
defn ref (text: String) -> Ref|False :
  if text == "false" :
    false
  else :
    ref-internal(text)

defn ref-internal (text: String) -> Ref :
  ;regex-match implementation incomplete so can do exact matches with "^mypattern$"
  ;val match-result = regex-match("^(\\w+).([\\w\.])+$", text)
  if text[length(text) - 1] == ']' :
    val idx = last(text, '[') as Int
    IndexRef(ref-internal(text[0 to idx]), to-int!(text[(idx + 1) to (length(text) - 1)]))
  else :
    val idx = last(text, '.')
    match(idx: Int) :
      FieldRef(ref-internal(text[0 to idx]), Ref(text[(idx + 1) to length(text)]))
    else :
      Ref(text)
```

utils/stm.stanza
```diff
@@ -221,26 +306,26 @@ defn create-pin-properties (json1:JSON) -> STMPinProperties :
...
-    for pin-item in rows(pin-properties) do:
-      println("%_, %_, %_, %_, %_" % [pin-name(pin-item), pad-data(pin-item), side(pin-item), has-generic-props?(pin-item), has-power-props?(pin-item)])
-
-    val stm-supports = create-supports(input-json)
-    println("Here")
-
-main()
\ No newline at end of file
+public defn create-mcu-struct (file:String) -> STM :
+  val input-json = parse $ slurp(file)
```
PhilippeFerreiraDeSousa, line 321: You need to add the necessary properties to MicroController in micro-controller.stanza,
adapt `public defn MicroController (json: JObject) -> MicroController :` and then adapt with the new properties (I don't think that the current version is used)
```
defmethod to-esir (mcu: MicroController) -> Instantiable :
  pcb-component my-microcontroller :
    ...
```

Then you delete `find-stm32-module` and
```
public defn micro-controller (params:Tuple<KeyValue>) :
  val mcu = MicroController(params)
  find-stm32-module(mpn(mcu))
```
becomes
```
public defn micro-controller (params:Tuple<KeyValue>) :
  val mcu = MicroController(params)
  to-esir(mcu)
```

This is the interface for database parts. It is the same with capacitors:
```
public defn Capacitor (raw-json: JObject) -> Capacitor :`
```

```
defmethod to-esir (capacitor: Capacitor) -> Instantiable :
  pcb-component my-capacitor :
```

```
public defn ceramic-cap (params:Tuple<KeyValue>) :
  val default-params = ["type" => "ceramic" 
                        "mounting" => "smd" 
                        "case" => get-valid-pkg-list()
                        "minimum_quantity" => 1
                        "min-rated-voltage" => MIN-CAP-VOLTAGE
                        "min-stock" => 5 * DESIGN-QUANTITY]
  to-esir(Capacitor(to-tuple $ to-hashtable<String, ?> $ cat(default-params params)))
```

Capacitor, inductor, Resistor, MicroController derive from Component
```
public deftype Component
public defmulti x (component: Component) -> Double
public defmulti y (component: Component) -> Double
public defmulti area (component: Component) -> Double
public defmulti to-esir (component: Component) -> Instantiable
public defmulti generate-esir (component: Component) -> Instantiable

defmethod area (component: Component) -> Double :
  x(component) * y(component)

defmethod generate-esir (cmp: Component) -> Instantiable :
  val esir-cmp = to-esir(cmp)
  ; ensure-generated(esir-cmp)
  esir-cmp
```
(I made `generate-esir`, in esir1 and didn't know how to adapt it to esir2 so it is useless)

PR #116, opened 2021-09-15
JITX-780: Added parameter values for soldermask opening and bridging

bpellenard:


PhilippeFerreiraDeSousa (approved):
Merge this before the CI God changes its mind!

m-hilgendorf (commented):
Do all these manufacturers have the same values?

manufacturers/rules.stanza
```diff
@@ -20,6 +20,8 @@ public pcb-rules bay-area-circuits-std-rules :
   min-silkscreen-text-height = 0.75 ; 29.5 mil (guess)
   solder-mask-registration = 0.106 ; 4 mil
   min-th-pad-expand-outer = 0.2032 ; 8 mil
+  min-soldermask-opening = 0.152
```
m-hilgendorf, line 23: Add `; TODO: lookup actual value` or something in here so we don't forget to update these with real values, otherwise we'll have designs pass verification that not invalid!

m-hilgendorf (changes requested):
In `design-vars.stanza` delete `MIN-SOLDER-MASK-SLIVER` and replace its usage in `checks.stanza`:

```
;utils/checks.stanza line 337
#CHECK(min-space(s) >= clearance(MinSolderMaskBridging, current-rules()))`
```

Since this change obsoletes the design var

bpellenard:
> Do all these manufacturers have the same values?

They are the values we discussed with @electricfungi

manufacturers/rules.stanza
bpellenard: It should be the real values

m-hilgendorf: It is incorrect for at least JLCPCB

https://jlcpcb.com/capabilities/Capabilities

bpellenard: Thanks

m-hilgendorf (approved):

PR #119, opened 2021-09-20
added paste layer to SMT land-patterns excluding test points.

electricfungi:
added paste layer to SMT land-patterns excluding test points.


m-hilgendorf (approved):
Looks good!

PR #114, opened 2021-09-14
Feature/jitx 522 stm to jitx

ariel-hi:
===== Microcontrollers in OCDB =====
Added new stm land patterns.
Rename to-esir to to-jitx.
Connect the new STM JSON to micro-controllers.stanza.
Generate JITX pin-properties, bundles, and supports dynamically.
Includes multiple test JSON files in the utils folder. These will be shuffled around.

(Recreating this PR from a newly named branch)

PR #124, opened 2021-09-21
use docker image jitx-client:0.9.0-dev.5 in CI

PhilippeFerreiraDeSousa:
Use a specific tag in CI


PhilippeFerreiraDeSousa:
Done in https://github.com/JITx-Inc/open-components-database/pull/114

PR #126, opened 2021-09-22
Cleanup ocdb/micro-controllers

m-hilgendorf:


utils/micro-controllers.stanza
```diff
@@ -145,47 +142,30 @@ defmethod to-jitx (mcu: MicroController) -> Instantiable :
 
     assign-landpattern $
       switch(mfg-package(mcu)) :
-        ; "SO8N" : SO8N
```
bpellenard, line 148: why?

bpellenard (approved):
Looks good to me!

utils/micro-controllers.stanza
m-hilgendorf: We can easily add it back once implemented in the database, this just removes the commented out code

PR #127, opened 2021-09-23
Added a simple unit test for land patterns

m-hilgendorf:
This is a simple sanity check applied to internal land patterns : 

- Is there a courtyard
- Is there a reference label
- Is there a polarity marker, if necessary


bpellenard (approved):
Looks good to me!

PR #129, opened 2021-09-24
Remove default params in query that are empty lists - temporarily remove octopart params

PhilippeFerreiraDeSousa:
I think I need to see what is going on with the octopart logic in the backend when stock != 0, in the mean time we can use stock=0 by default which won't call the octopart logic by default (thanks to jitx-client PR https://github.com/JITx-Inc/jitx-client/pull/356 that removes default sellers).

This will also remove the stock requirements on resistors / inductors / capacitors which was useful.
Will revert default stock when I find out if something is off with the octopart logic, will probably reduce it as 100 is small for resistors but a lot for mcus.

### DIFF
--- a/designs/mcu.stanza
+++ b/designs/mcu.stanza
@@ -1,5 +1,5 @@
 #use-added-syntax(jitx)
-defpackage ocdb/designs/tutorial :
+defpackage ocdb/designs/mcu :
   import core
   import collections
   import jitx
--- a/utils/micro-controllers.stanza
+++ b/utils/micro-controllers.stanza
@@ -107,7 +107,7 @@ public defn MicroController (properties:Tuple<KeyValue>, exist:Tuple<String>, so
   if DESIGN-QUANTITY > 0 :
     add(octopart-parameters, "_stock" => DESIGN-QUANTITY)
   match(vendors: Tuple<String>) :
-    if length(vendors) != 0 :
+    if not empty?(length(vendors)) :
       add(octopart-parameters, "_sellers" => vendors)
 
   val design-parameters = Vector<KeyValue<String, ?>>()
--- a/utils/micro-controllers.stanza
+++ b/utils/micro-controllers.stanza
@@ -105,8 +105,9 @@ public defn MicroController (user-properties:Tuple<KeyValue>, exist:Tuple<String
   val optional-properties =
     generate<KeyValue<String, ?>> :
       ; Octopart parameters: if they exist, octopart will be queried
-      yield("_stock" => DESIGN-QUANTITY) when DESIGN-QUANTITY > 0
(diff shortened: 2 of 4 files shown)


utils/micro-controllers.stanza
```diff
@@ -97,33 +97,34 @@ public defn MicroController (properties:Tuple<KeyValue>) -> MicroController :
...
-                                            "_stock" => DESIGN-QUANTITY],
-                                          vendor-properties
-                                          temperature-properties,
-                                          properties])
+
+  val octopart-parameters = Vector<KeyValue<String, ?>>()
+  if DESIGN-QUANTITY > 0 :
+    add(octopart-parameters, "_stock" => DESIGN-QUANTITY)
+  match(vendors: Tuple<String>) :
+    if length(vendors) != 0 :
```
bpellenard, line 110: `empty?`?

bpellenard (approved):
Looks good to me!

utils/micro-controllers.stanza
```diff
@@ -97,33 +97,34 @@ public defn MicroController (properties:Tuple<KeyValue>) -> MicroController :
...
-  else :
-    query-properties = to-tuple $ cat-all([["category" => "microcontroller",
-                                            "_sort" => sort,
-                                            "_exist" => exist,
-                                            "_stock" => DESIGN-QUANTITY],
-                                          vendor-properties
-                                          temperature-properties,
-                                          properties])
+
+  val octopart-parameters = Vector<KeyValue<String, ?>>()
```
m-hilgendorf, line 106: perfect place for a (stanza) generator. 
```
val optional-properties =
  generate<KeyValue> : 
    yield("_stock" => DESIGN-QUANTITY) when DESIGN-QUANTITY > 0
    yield("_sellers" => vendors) when call?<Tuple>(empty?, vendors) is-not False ; import lang-utils
    yield("_sort" => sort) when not empty?(sort)
    yield("_exist" => exist) when not empty?(exist)
    match(operating-temperature:[Double, Double]):
      yield("max-rated-temperature.min" => operating-temperature[0])
      yield("min-rated-temperature.max" => operating-temperature[1])

  val query-properties = to-tuple(cat-all([properties, optional-properties]))
  val result = dbquery-all(query-properties) as JObject
  throw(NoComponentMeetingRequirements(query-properties)) when empty?(result)
  result 
```

PhilippeFerreiraDeSousa: Cool trick the `call?`

designs/mcu.stanza
```diff
@@ -1,5 +1,5 @@
 #use-added-syntax(jitx)
-defpackage ocdb/designs/tutorial :
+defpackage ocdb/designs/mcu :
```
ariel-hi, line 2: 👍

ariel-hi (approved):
Awesome

PR #130, opened 2021-09-24
Stm error handling

m-hilgendorf:
`fatal(...)` calls in the STM extractor now throw exceptions, allowing for recovery in higher level generators. 

Pending support for multiple pads mapped to the same pin, an error is introduced to handle this case:

```
FATAL ERROR: Multiple pins are mapped to the same pad:
  pad 19: PA[9], PA[11]
  pad 20: PA[10], PA[12].
```

---

The component generation is now cacheable, via converting the STMxxx structs to pcb-structs

---

Land pattern lookup is cleaned up and comprehensive using a table of symbols and the land pattner name directly. 

--- 

Integer pins are converted from doubles earlier in extraction to avoid confusion during printout of errors.

### DIFF
--- a/components/st-microelectronics/landpatterns.stanza
+++ b/components/st-microelectronics/landpatterns.stanza
@@ -257,3 +257,55 @@ public pcb-landpattern UFQFPN48:
                       tol(0.4, 0.1),   ;Lead size in x direction:   L
                       tol(0.25, 0.05), ;Lead size in y direction:   b
                       GeneralChamferedRectangle(5.6, 5.6, [0.5 0. 0. 0.]))
+
+public val STM32-LAND-PATTERNS = [
+  UFBGA64
+  UFBGA132
+  LFBGA100
+  LFBGA144
+  LFBGA354
+  TFBGA100
+  TFBGA216
+  TFBGA225
+  TFBGA64
+  UFBGA100
+  UFBGA121
+  UFBGA129
+  UFBGA144
+  UFBGA169
+  UFBGA73
+  WLCSP18
+  WLCSP20
+  WLCSP25
+  WLCSP36
+  WLCSP49
+  WLCSP52
+  WLCSP63
+  WLCSP64
+  WLCSP66
+  WLCSP72
+  WLCSP81
+  WLCSP90
+  WLCSP100
+  WLCSP104
+  WLCSP115
+  WLCSP132
+  WLCSP143
+  WLCSP144
+  WLCSP156
+  WLCSP168
+  WLCSP180
+  TSSOP20 
+  LQFP100
+  LQFP128
+  LQFP144
+  LQFP176
+  LQFP208
+  LQFP32
+  LQFP48
+  LQFP64
+  LQFP80
+  UFQFPN28
+  UFQFPN32
+  UFQFPN48
+]
--- a/utils/micro-controllers.stanza
+++ b/utils/micro-controllers.stanza
@@ -62,6 +62,7 @@ public defstruct MicroController <: Component :
   bundles: Tuple<STMBundle>
   supports: Tuple<STMSupports>
 
+
 public defn MicroController (json: JObject) -> MicroController :
   val [x, y, z] = parse-dimensions(json["dimensions"] as JObject)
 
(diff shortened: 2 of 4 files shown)


components/st-microelectronics/landpatterns.stanza
```diff
@@ -257,3 +257,55 @@ public pcb-landpattern UFQFPN48:
                       tol(0.4, 0.1),   ;Lead size in x direction:   L
                       tol(0.25, 0.05), ;Lead size in y direction:   b
                       GeneralChamferedRectangle(5.6, 5.6, [0.5 0. 0. 0.]))
+
+public val STM32-LAND-PATTERNS = [
```
m-hilgendorf, line 261: This table includes all the land patterns in this file.

utils/micro-controllers.stanza
```diff
@@ -130,68 +131,20 @@ public defn MicroController (properties:Tuple<KeyValue>, exist:Tuple<String>, so
...
-        "WLCSP66" : WLCSP66
-        "WLCSP72" : WLCSP72
-        "WLCSP81" : WLCSP81
-        "WLCSP90" : WLCSP90
-        else :
-          fatal("MCU package %_ is not supported." % [mfg-package(mcu)])
-    make-box-symbol()
-  my-microcontroller
-
+pcb-component mcu-component (manu:String, pn:String, pp:STMPinProperties, supports:Tuple<STMSupports>, bundles:Tuple<STMBundle>, pkg:String) :
```
m-hilgendorf, line 134: The component generation can now be cached

utils/micro-controllers.stanza
```diff
@@ -130,68 +131,20 @@ public defn MicroController (properties:Tuple<KeyValue>, exist:Tuple<String>, so
...
+  manufacturer = manu
+  mpn = pn
+  to-jitx-pin-properties(pp)
+  ; println(pp)
+  val bundle-table = to-jitx-bundles(bundles)
+  to-jitx-supports(supports, bundle-table)
+  val lp = find({name(_) == pkg}, STM32-LAND-PATTERNS)
+  throw(Exception("%_ is not a supported package." % [pkg])) when lp is False
```
m-hilgendorf, line 141: Handle the case where the database returns a component without a matching land pattern explicitly

utils/stm-to-jitx.stanza
```diff
@@ -33,7 +33,7 @@ public defn to-jitx-pin-properties (pin-properties:STMPinProperties) -> False :
     val power-props = stm-to-jitx $ power-pin(pin-properties)
     
     ;Determine if the pads numeric (Double -> Int) or named (Ref -> String).
-    val numeric-pads? = any?({pad(_) is Double}, rows(pin-properties))
+    val numeric-pads? = any?({pad(_) is Int}, rows(pin-properties))
```
m-hilgendorf, line 36: numeric pads are integers, so we convert earlier

components/st-microelectronics/landpatterns.stanza
electricfungi: This is not something we should support - I'm not aware of any situation where multiple pins should be mapped to the same pad

electricfungi: ok chatted with mike this is middleware and at the schematic symbol level 1 pad will not be mapped to more than 1 schematic symbol pin

utils/stm.stanza
```diff
@@ -6,58 +7,72 @@ defpackage ocdb/stm:
   import reader
   import jitx
 
+defstruct StmError <: Exception : 
+  message:?
```
bpellenard, line 11: message could be anything, right? Do not we want to be more restrictive?

utils/stm.stanza
```diff
@@ -6,58 +7,72 @@ defpackage ocdb/stm:
...
 
 ;JITX Pin Properties to be created in JITX
-public defstruct STMPinProperties :
+public pcb-struct ocdb/stm/STMPinProperties :
   generic-pin: STMGenericPin
   power-pin: STMPowerPin
   rows: Tuple<STMPinProperties-Row>
 
-public defstruct STMGenericPin :
+public pcb-struct  ocdb/stm/STMGenericPin :
```
bpellenard, line 34: you have two spaces instead of one between `pcb-struct` and `ocdb/...`

utils/stm.stanza
```diff
@@ -129,6 +144,8 @@ defmethod print (o:OutputStream, s:STMSupportMapping) :
 ;=================================================
 ;================= Build Structs =================
 ;=================================================
+defn stm-error (x) :
```
bpellenard, line 147: exception would be more appropriate

utils/stm.stanza
m-hilgendorf: This allows adding more detail to the type incrementally, right now it's just String|Printable

utils/stm.stanza
m-hilgendorf: Fixed!

utils/stm.stanza
m-hilgendorf: Fixed!

utils/stm.stanza
m-hilgendorf: Fixed, and made public

utils/micro-controllers.stanza
ariel-hi: Quite genius!

ariel-hi: Perhaps instead of `manu` and `pn` which aren't immediately clear:
`mcu-manufacturer` and `mcu-mpn`

m-hilgendorf: good point, fixed

utils/stm-to-jitx.stanza
```diff
@@ -42,31 +43,35 @@ public defn to-jitx-pin-properties (pin-properties:STMPinProperties) -> False :
     if not numeric-pads? and not named-pads? :
       fatal("Generated a micro-controller without pads.") 
     
+    ; We need to compile a list of reused pad to pin mappings.
+    val pad-mappings = HashTable<String|Int, Vector<String>>()
+    defn check-pad (pad:String|Int, pin:String):
+      if get?(pad-mappings, pad) is False : 
+        pad-mappings[pad] = Vector<String>()
+      add(pad-mappings[pad], pin)
+      length(pad-mappings[pad]) == 1
```
ariel-hi, line 52: You can set a default vector for a table. This lets you add a value without checking if the entry already exists.
```
    val pad-mappings = HashTable-init<String|Int, Vector<String>>(Vector<String>{})
    defn check-pad (pad:String|Int, pin:String) -> True|False :
      add(pad-mappings[pad], pin)
      length(pad-mappings[pad]) == 1
```

utils/stm-to-jitx.stanza
```diff
@@ -105,26 +113,25 @@ public defn to-jitx-supports (supports:Tuple<STMSupports>, bundle-table:HashTabl
       supports b1 :
         option :
           for mapping in mappings do :
+            println("mapping:%_" % [mapping])
```
ariel-hi, line 116: Leftover print statement

utils/stm-to-jitx.stanza
m-hilgendorf: Good trick, thanks!

utils/stm-to-jitx.stanza
m-hilgendorf: Fixed

utils/stm-to-jitx.stanza
m-hilgendorf: Using 

```
    val pad-mappings = HashTable<String|Int, Vector<String>>(Vector<String>())
```

Caused a lookup error, so I'm leaving the original for now.

ariel-hi: Sure, yours works too. In the future make sure to use curlies `...(Vector<String>{})`

utils/stm-to-jitx.stanza
```diff
@@ -42,30 +43,36 @@ public defn to-jitx-pin-properties (pin-properties:STMPinProperties) -> False :
...
-          val pad-name = raw-string-to-ref(pad(row) as String)
+    pin-properties : 
+      [pin:Ref | pads : (Ref|Int) ... | side:Dir | generic-pin:GenericPin | power-pin:PowerPin]
+      for row in rows(pin-properties) do : 
+        val pin-ref = ref-string-to-ref(pin(row))
+        val pad-ref = match(pad(row)):
+          (i:Int): i
+          (s:String): raw-string-to-ref(s)
+
+        if check-pad(pad(row), pin(row)) :
```
PhilippeFerreiraDeSousa, line 64: So we silently ignore when several pins are mapped to the same pad? Why does this happen in the first place?

ariel-hi (approved):
Great, thanks for the changes and updates.

utils/stm-to-jitx.stanza
m-hilgendorf: We're not ignoring it, it's possible for multiple pin names to refer to the same schematic pin. We pick the first we come across as the canonical pin. 

Ideally we would rename, but that is a bit trickier given the naming conventions so we can fix that in a subsequent PR

PR #138, opened 2021-09-28
Fix LFBGA354 and WLCSP72 land patterns

m-hilgendorf:
LFBGA354:

(notice: mirrored)

lfbga354-doc
lfbga354-corrected

WLCSP72:
wlcsp72-doc
wlcsp72


PhilippeFerreiraDeSousa (dismissed):
If you checked it is correct, then it looks good to me

bpellenard:
Are the overlaps ok on the second picture?

m-hilgendorf:
That's controlled by the solder mask registration design rule, which is too large by default for this land pattern. It is not controlled in this source code.

bpellenard (changes requested):
The first one is wrong, there is a gap of two in the rows.

components/st-microelectronics/landpatterns.stanza
```diff
@@ -29,9 +29,19 @@ public pcb-landpattern LFBGA144:
...
 public pcb-landpattern LFBGA354:
-  defn depop? (r:Int, c:Int):
-    (c == 19 and contains?([5, 15], r)) or
-    (c == 16 and contains?([6, 8, 10, 12, 14], r))
+  defn depop? (r:Int, c:Int) :
+    contains?(
+      [
+        [4  18],
+        [14 18],
+        [5  17],
```
bpellenard, line 37: [5, 15]

m-hilgendorf: Fixed!

m-hilgendorf:
Good catch, fixed
lfbga354-fixed
:

bpellenard (approved):
Looks good to me!

PR #139, opened 2021-09-28
Comprehensive landpattern generation

m-hilgendorf:
Adds two variants to `make-two-pin-chip-landpattern` to mirror the comprehensiveness of `make-xxx` and `xxx` generator families. 

These are useful when you wish to use the form

```
pcb-landpattern MyPassive : 
  make-two-pin-chip-landpattern(...)
```

as opposed to 

```
val MyPassive = two-pin-chip-landpattern(...)
```


bpellenard (approved):
Looks good to me!

PR #147, opened 2021-10-04
extended get-valid-pkg-list to accept optional package

electricfungi:
extended get-valid-pkg-list to accept optional package, which in turn adds function to specify a min package for a specific instance of a generic component


PhilippeFerreiraDeSousa:
Thank you, this will conflict with https://github.com/JITx-Inc/open-components-database/pull/146/files

RajeshJITX (commented):
Is the intent of this PR to allow get-valid-pkg-list call with different parameters other than MIN-PKG?

electricfungi:
Resolved with PR 146

PR #148, opened 2021-10-04
look-up-micro-controller

PhilippeFerreiraDeSousa:
Create new function to look up what are the possible values that an attribute can take taking into consideration the same settings as the `micro-controller` function


ariel-hi (approved):
This is useful!

PR #150, opened 2021-10-05
Add back changes from release-0.9.0 banches

PhilippeFerreiraDeSousa:


.github/jitx-client/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM 657302324634.dkr.ecr.us-west-2.amazonaws.com/jitx-client:0.9.0-dev.7
+FROM 657302324634.dkr.ecr.us-west-2.amazonaws.com/jitx-client:0.9.0-rc.4
```
PhilippeFerreiraDeSousa, line 1: rc.4 is sligthly more recent than dev.7 so it is good

m-hilgendorf (approved):

PR #178, opened 2021-11-10
Improvements to the STM api including more setup options

bhusang:
Further optimizing of STM api to address usability, readability
1) more bypass options (separate package and pin calls) with backwards compatibility
2) added more enums for vdd, reset, boot
3) tweaked default settings
4) added some options for battery connection
5) added options for reset/por


d-haldane:
@bhusang The `generate-pin-bypass` function is probably not doing what you want. Let's say we have `VDD[3]` as a port on the microcontroller, I would expect `generate-pin-bypass` to generate 3 bypass caps and associate them with `VDD[0]` `VDD[1] and `VDD[2]`, but the function only puts one down for the whole port.

d-haldane:
In stm-to-jitx, you need to have one version of the function (like `connect-reset`) that passes in _all_ the relevant parameters from the settings.

Right now, I can set `reset-pullup => 100.0e3 and it just gets ignored. 

So 
```
connect-power
connect-reset
connect-por
connect-battery
generate-bypass
generate-pin-bypass
set-boot
connect-debug
setup-clocks
```
Need to have a single call in `stm-to-jitx` that passes through all the information you need to stm-api to generate the correct circuitry. stm-to-jitx will be called the same way on every single MCU, so you don't get to do things like `connect-power(mcu, "VDDIO2", VDDIO2, VSS)` because most MCUs will not have that pin.

bhusang:
@d-haldane I left the full-featured api calls as comments in stm-to-jitx. I will uncomment the full function call instead. I will remove the less useful function calls to make it cleaner.

bhusang:
Given that some micros will have VDDIO2 so if we are calling the same stm-to-jitx function on the whole set, we need someway to support those that have them and those that don't. I made the function call not fail if there is no associated pin.

bhusang:
The generate-pin-bypass was intended to support a single pin, but you example demonstrates the issue with that logic. I will modify the function to always add capacitors to multi-pin nets.

bhusang:
fixed https://github.com/JITx-Inc/open-components-database/pull/178#issuecomment-965646986
addressed https://github.com/JITx-Inc/open-components-database/pull/178#issuecomment-965659856

components/st-microelectronics/stm-api.stanza
```diff
@@ -130,62 +137,72 @@ defn match-pin? (pin-ref:Ref power:String) :
...
-      match-pin?(ref(p), my-pin)
+; find the pin exact match return a boolean
+defn find-pin? (obj:JITXObject my-pin:String) :
+  val result = get-pin-list(obj, my-pin)
   not empty?(result)
 
-defn power-pins (stm:JITXObject vdd-pin:PowerPinNames) -> Tuple<JITXObject> :
-  val found-pins = find-pins-list(stm, to-string(vdd-pin))
+defn power-pins (stm:JITXObject vdd-pin:PwrPinNames) -> Tuple<JITXObject> :
+  val found-pins = get-pin-list(stm, to-string(vdd-pin))
```
m-hilgendorf, line 153: Can we add some context to the error message? This can help us when we debug: 

```
fatal("%_ is missing an identifiable power pin (%_)" % [ref(stm), vdd-pin)])
```

components/st-microelectronics/stm-api.stanza
```diff
@@ -208,54 +224,75 @@ public defn connect-named-power (stm:JITXObject, power-port-name:String, vdd-net
...
+        net (b-net p-pin[0])
 
+public defn connect-battery (stm:JITXObject, bat-net:PwrPinNames) :
+  connect-battery(stm, bat-net, VDD)
 
-public defn connect-reset (stm:JITXObject) :
+public defn connect-battery (stm:JITXObject) :
+  connect-battery(stm, VBAT, VDD)
+
+public defn connect-reset (stm:JITXObject reset-net:ResetPinNames vdd-net:PwrPinNames gnd-net:GndPinNames reset-pullup:Double reset-cap:Double) :
```
electricfungi, line 245: I like the idea of extending this! But we could go further, enable reset-pullup and reset-cap to be JITXObjects, then users can specify things like low leakage caps or packages, or things we may not even think about today.

bhusang: @electricfungi So I think that's a good idea. I tried the following (but it did not like the fact that I was calling the JITXObject as a function):
```
public defn connect-reset (stm:JITXObject reset-net:ResetPinNames vdd-net:PwrPinNames gnd-net:GndPinNames reset-pullup:JITXObject|Double reset-cap:JITXObject|Double) :
  inside pcb-module :
    pin reset
    val reset-pin-list = get-pin-list(stm, reset-net)
    if not empty?(reset-pin-list) :
      val power-pins = power-pins-list(stm, VDD)
      for p in reset-pin-list do :
        net (p reset)
        if not empty?(power-pins) :
          match(reset-pullup) :
            (r: Double)     : res-strap(power-pins[0], p, r)
            (i: JITXObject) : inst rp : i(power-pins[0], p)
          match(reset-cap) :
            (c: Double)     : cap-strap(power-pins[0], p, c)
            (i: JITXObject) : inst cp : i(power-pins[0], p)
```
How should I fix this -- am I correct in the approach?

bhusang: I have tried the following but the Instantiable cannot take any arguments.
```
public defn connect-reset (stm:JITXObject reset-net:ResetPinNames vdd-net:PwrPinNames gnd-net:GndPinNames reset-pullup:Instantiable|JITXObject|Double reset-cap:Instantiable|JITXObject|Double) :
  inside pcb-module :
    pin reset
    val reset-pin-list = get-pin-list(stm, reset-net)
    if not empty?(reset-pin-list) :
      val power-pins = power-pins-list(stm, VDD)
      for p in reset-pin-list do :
        net (p reset)
        if not empty?(power-pins) :
          match(reset-pullup) :
            (r: Double)       : res-strap(power-pins[0], p, r)
            (i: Instantiable) : inst rp : i
            (j: JITXObject)   : false
          match(reset-cap) :
            (c: Double)     : cap-strap(power-pins[0], p, c)
            (i: Instantiable) : inst cp : i
            (j: JITXObject)   : false
```

bhusang: @electricfungi @CuppoJava Here is my evolution of the reset-net function. Note that it still does not cover that many cases.
```
public defn connect-reset (stm:JITXObject reset-net:ResetPinNames vdd-net:PwrPinNames gnd-net:GndPinNames reset-pullup:Instantiable|JITXObject|Double reset-cap:Instantiable|JITXObject|Double) :
  inside pcb-module :
    pin reset
    val reset-pin-list = get-pin-list(stm, reset-net)
    if not empty?(reset-pin-list) :
      val power-pins = power-pins-list(stm, VDD)
      for p in reset-pin-list do :
        net (p reset)
        if not empty?(power-pins) :
          match(reset-pullup) :
            (r: Double)       : res-strap(power-pins[0], p, r)
            (i: Instantiable) : let :
                                  inst rp : i
                                  net (rp.p[0] power-pins[0])
                                  net (rp.p[1] p)
            (j: JITXObject)   : false
          match(reset-cap) :
            (c: Double)     : cap-strap(power-pins[0], p, c)
            (i: Instantiable) : let :
                                  inst cp : i
                                  net (cp.p[0] power-pins[0])
                                  net (cp.p[1] p)
            (j: JITXObject)   : false
```

components/st-microelectronics/stm-api.stanza
```diff
@@ -130,132 +138,177 @@ defn match-pin? (pin-ref:Ref power:String) :
   if p-ref == power :
     true
 
-defn find-pins-list (obj:JITXObject my-pin:String) :
+; return list of pins that match name (exactly) / don't check for empty
+defn get-pin-list (obj:JITXObject pin-id:?) :
```
ariel-hi, line 142: Can you separate function parameters with a comma?
It's perfectly valid without, but it's an unspoken standard.
`(obj:JITXObject, pin-id:?)`

bhusang: Oh, ok will do.

components/st-microelectronics/stm-api.stanza
```diff
@@ -130,132 +138,177 @@ defn match-pin? (pin-ref:Ref power:String) :
...
+    if substring?(to-string(ref(p)), pin-name) :
+      match-pin?(ref(p), pin-name)
   result
 
-defn find-pins? (obj:JITXObject my-pin:String) :
-  val result = to-tuple $ for p in pins(obj) filter :
-    if substring?(to-string(ref(p)), my-pin) :
-      match-pin?(ref(p), my-pin)
+; find the pin exact match return a boolean
+defn find-pin? (obj:JITXObject my-pin:String|PwrPinNames|GndPinNames|ResetPinNames|BootPinNames|ReferencePinNames) :
```
ariel-hi, line 150: I recommend creating a new type that each PinNames subtypes.
```
public deftype STMPinNames
public pcb-enum .../PwrPinNames <: STMPinNames
...
```

```
defn get-pin-list (obj:JITXObject pin-id:String|STMPinNames)
defn find-pin? (obj:JITXObject my-pin:String|STMPinNames)
```

components/st-microelectronics/stm-api.stanza
```diff
@@ -130,132 +138,177 @@ defn match-pin? (pin-ref:Ref power:String) :
...
+  val result = get-pin-list(obj, my-pin)
   not empty?(result)
 
-defn power-pins (stm:JITXObject vdd-pin:PowerPinNames) -> Tuple<JITXObject> :
-  val found-pins = find-pins-list(stm, to-string(vdd-pin))
+defn power-pins (stm:JITXObject vdd-pin:PwrPinNames) -> Tuple<JITXObject> :
+  val found-pins = get-pin-list(stm, vdd-pin)
   if empty?(found-pins) :
-    fatal("%_ is missing an identifiable power pin." % [ref(stm)])
+    fatal("%_ is missing an identifiable power pin (%_)." % [ref(stm), to-string(found-pins[0])])
```
ariel-hi, line 157: This will break --
`found-pins` is empty so its first item cannot be accessed.

components/st-microelectronics/stm-api.stanza
```diff
@@ -130,132 +138,177 @@ defn match-pin? (pin-ref:Ref power:String) :
...
 
+defn power-pins-list (stm:JITXObject vdd-pin:PwrPinNames|ReferencePinNames) -> Tuple<JITXObject> :
+  get-pin-list(stm, vdd-pin)
+
 defn gnd-pins (stm:JITXObject, gnd-pin:GndPinNames) -> Tuple<JITXObject> :
-  val found-pins = find-pins-list(stm, to-string(gnd-pin))
+  val found-pins = get-pin-list(stm, gnd-pin)
   if empty?(found-pins) :
-    fatal("%_ is missing an identifiable ground pin." % [ref(stm)])
+    fatal("%_ is missing an identifiable ground pin (%_)." % [ref(stm), to-string(found-pins[0])])
```
ariel-hi, line 169: Likewise, this will break.

components/st-microelectronics/stm-api.stanza
```diff
@@ -130,132 +138,177 @@ defn match-pin? (pin-ref:Ref power:String) :
...
   inside pcb-module :
-    val power-pin-list = power-pins(stm, VDD)
-    val gnd-pin-list = gnd-pins(stm, VSS)
-    bypass-cap-strap(power-pin-list[0], gnd-pin-list[0], bypass-package)
-    for p in power-pin-list do :
-      bypass-cap-strap(p,  gnd-pin-list[0], bypass-pin)
+    val power-pins = power-pins-list(stm, vdd-net)
+    val gnd-pins = gnd-pins-list(stm, gnd-net)
+    if not empty?(power-pins) and not empty?(gnd-pins) :
+      bypass-cap-strap(power-pins[0], gnd-pins[0], cap-bypass-package)
```
ariel-hi, line 198: This is clean!
Should we inform the user if they requested a package bypass but one couldn't be created?

components/st-microelectronics/stm-api.stanza
```diff
@@ -130,132 +138,177 @@ defn match-pin? (pin-ref:Ref power:String) :
...
+        val v-range = recommended-voltage(property(vdd-pin-list[0].power-pin))
+        if contains?(v-range, 3.3) :
+          property(vdd-pin-list[0].power-request) = [3.3, 0.1, 3.3 * 0.03]
+        else if contains?(v-range, 2.5) :
+          property(vdd-pin-list[0].power-request) = [2.5, 0.1, 2.5 * 0.03]
+        else if contains?(v-range, 1.8) :
+          property(vdd-pin-list[0].power-request) = [1.8, 0.1, 1.8 * 0.03]
+        else if contains?(v-range, 1.2) :
+          property(vdd-pin-list[0].power-request) = [1.2, 0.1, 1.2 * 0.03]
+        true ; need this in order to check for missing pins in the outer if statements
```
ariel-hi, line 229: Great job creating nets!

components/st-microelectronics/stm-api.stanza
```diff
@@ -125,137 +133,181 @@ defn parse (my-ref: Ref) -> String :
...
+
+public defn connect-por (stm:JITXObject, por-net:ResetPinNames, vdd-net:PwrPinNames, gnd-net:GndPinNames) :
+  inside pcb-module :
+    pin por
+    val por-pin = get-pin-list(stm, por-net)
+    if not empty?(por-pin) :
+      val power-pins = power-pins-list(stm, vdd-net)
+      val gnd-pins = gnd-pins-list(stm, gnd-net)
+      val v-range = recommended-voltage(property(power-pins[0].power-pin))
+      if not empty?(power-pins) and not empty?(gnd-pins) :
```
ariel-hi, line 298: Perhaps `v-range` should be calculated after knowing `power-pins` is not empty?

components/st-microelectronics/stm-api.stanza
bhusang: Ah, good catch

components/st-microelectronics/stm-api.stanza
bhusang: I think that the user will have bigger issues if the bypass is not able to be created -- the connect-power(stm, "power", VDD, VSS) will fail hard. But I guess that more info is useful. Will add, thanks.

bhusang: Actually, the function will return false if no bypass-cap-strap is instantiated. We could always look at the return value for the function?

ariel-hi (changes requested):
Great! I recommended a few changes in stm-api.
Functions like `match-pin?` and `get-pin-list` would benefit from unit testing.

components/st-microelectronics/stm-api.stanza
ariel-hi: Yes, the `false` is a good indicator already! Good point

components/st-microelectronics/stm-api.stanza
bhusang: This is a great suggestion for readability. Done.

components/st-microelectronics/stm-api.stanza
bhusang: This was actually a bug -- the found-pins[0] should have been gnd-net. But thanks for the pointer.

components/st-microelectronics/stm-api.stanza
bhusang: Yes, good feedback. Changed.

.github/jitx-client/user.params
```diff
@@ -1,2 +1,6 @@
+server :
+  base-url = "https://app-testing.jitx.com"
+  url = "https://app-testing.jitx.com/api/"
+
```
ariel-hi, line 4: This points to the testing server, so it should be removed.

bhusang: I believe that it was added to allow the CI to run successfully. The normal CI fails on an error that is pending.

bhusang: I can remove it but the CI will fail when I do the push.

bhusang: And then the PR can't be merged :|

ariel-hi (approved):
Thanks for following up on changes! Great work

PR #179, opened 2021-11-10
Explorer run functions

PhilippeFerreiraDeSousa:
Creates `chip-resistors`, `smd-inductors`, `ceramic-capacitors` to be used in the explorer. This is copy pasting the singular versions of those functions but returning several results.
To do so change `Resistors`, `Inductors` and `Capacitors` to take a limit arg instead of `exist` which was there only because I used it once in the voltage divider 1 year ago but nobody uses it.

### DIFF
--- a/utils/design-vars.stanza
+++ b/utils/design-vars.stanza
@@ -14,7 +14,7 @@ public pcb-enum ocdb/design-vars/DensityLevel:
 public var OPERATING-TEMPERATURE = [0.0 25.0]
 
 ; Part selection parameters
-public var OPTIMIZE-FOR = ["area"]
+public var OPTIMIZE-FOR: Tuple<String> = ["area"]
 public var MAX-Z = 500.0
 public var MIN-PKG = "0201"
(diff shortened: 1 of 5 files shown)


utils/design-vars.stanza
```diff
@@ -14,7 +14,7 @@ public pcb-enum ocdb/design-vars/DensityLevel:
 public var OPERATING-TEMPERATURE = [0.0 25.0]
 
 ; Part selection parameters
-public var OPTIMIZE-FOR = ["area"]
+public var OPTIMIZE-FOR: Tuple<String> = ["area"]
```
PhilippeFerreiraDeSousa, line 17: Type hinting because the type was wrongly inferred to [String] while you should be able to assign a Tuple<String>:
```
tools/explorer.stanza:46.10: Cannot assign expression of type Tuple<String> to variable OPTIMIZE-FOR
with declared type [String].
```

m-hilgendorf (approved):
LGTM

PR #182, opened 2021-12-02
changed thermal pad issue from fatal() to print() with altered messaging

bhusang:


utils/land-patterns.stanza
```diff
@@ -1086,7 +1086,7 @@ defn make-thermal-pad (shape?:False|Shape, max-thermal-dims:Dims, num-pins:Int):
     val thermal-dims = dims(shape?)
     if x(thermal-dims) >= x(max-thermal-dims) or
        y(thermal-dims) >= y(max-thermal-dims):
-      fatal("make-thermal-pad(...) failed: shape exceeds maximum thermal pad dimensions. Called with %_, while the maximum dimensions are %_." % [
```
m-hilgendorf, line 1089: How should we handle courtyards without this check?

m-hilgendorf: For example say you make a thermal pad larger than the computed courtyard, do we want to fail still? Or incorporate the thermal pad dimensions into the courtyard computation?

bhusang: I would keep the courtyard as is for now. This check only disallowed the thermal pad from straying into the normal pad area but did nothing for courtyards. There is probably another discussion of how to refine the courtyard.
```
    val solder-mask-registration = clearance(current-rules(), SolderMaskRegistration)
    val part-sz = Dims(max-value(package-length), max-value(package-width))
    val lp-sz   = enlarge(Dims(Zmax(ipc-ew), Zmax(ipc-ns)), 2.0 * solder-mask-registration)
    val cy-sz   = enlarge(max(part-sz, lp-sz), courtyard-excess(lead-fillets))
    layer(Courtyard(Top)) = Rectangle(cy-sz)
```

m-hilgendorf: Ok. In that case, delete the check and print statement. I believe the pcb-landpattern checks are more comprehensive and will correctly report overlaps.

bhusang: @m-hilgendorf I have created a Linear issue (https://linear.app/jitx/issue/JITX-1026/courtyard-check-should-be-enhanced-to-include-all-component-objects) to add a better check on the courtyard size.

bhusang: Ok, removed the print, function arguments from the make-thermal-pad function

m-hilgendorf (approved):

PR #184, opened 2021-12-03
Tant caps

electricfungi:
Add tantalum cap functions. Fix issue with polarized cap symbol mapping

### DIFF
--- a/utils/db-parts.stanza
+++ b/utils/db-parts.stanza
@@ -788,15 +788,16 @@ defmethod to-jitx (capacitor: Capacitor) -> Instantiable :
     val esr = esr(capacitor)
     val case = case(capacitor)
     if pol? :
+      println("Got here, pol ")
       pin a
       pin c
-
       val sym = capacitor-sym(CapacitorPolarized)
       symbol = sym(a => sym.p[1], c => sym.p[2])
+      println("Case %_ " % [case] )
 
       match(case: String) :
         if all?(digit?, to-seq(case)) :
-          val pkg = ipc-two-pin-landpattern(case)
+          val pkg = ipc-two-pin-landpattern(case, pol?)
           landpattern = pkg(a => pkg.a, c => pkg.c)
         else :
           val pkg = dummy-landpattern(2, [x(capacitor) y(capacitor)])
--- a/utils/db-parts.stanza
+++ b/utils/db-parts.stanza
@@ -793,7 +793,6 @@ defmethod to-jitx (capacitor: Capacitor) -> Instantiable :
       pin c
       val sym = capacitor-sym(CapacitorPolarized)
       symbol = sym(a => sym.p[1], c => sym.p[2])
-      println("Case %_ " % [case] )
 
       match(case: String) :
         if all?(digit?, to-seq(case)) :
--- a/utils/db-parts.stanza
+++ b/utils/db-parts.stanza
@@ -788,7 +788,6 @@ defmethod to-jitx (capacitor: Capacitor) -> Instantiable :
     val esr = esr(capacitor)
     val case = case(capacitor)
     if pol? :
-      println("Got here, pol ")
       pin a
       pin c
       val sym = capacitor-sym(CapacitorPolarized)
--- a/utils/generic-components.stanza
+++ b/utils/generic-components.stanza
@@ -179,7 +179,7 @@ public defn look-up-ceramic-caps (attribute: String, filter-properties:Tuple<Key
 ; ========================================================
 ; ==================== tantalum-cap =======================
 ; ========================================================
-
+; Use for adding an instance of a tantalum cap from the JITX database to your design.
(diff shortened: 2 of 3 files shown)


utils/db-parts.stanza
```diff
@@ -788,15 +788,15 @@ defmethod to-jitx (capacitor: Capacitor) -> Instantiable :
     val esr = esr(capacitor)
     val case = case(capacitor)
     if pol? :
+      println("Got here, pol ")
```
m-hilgendorf, line 791: Remove print debugging

utils/generic-components.stanza
```diff
@@ -266,6 +176,41 @@ public defn look-up-ceramic-caps (attribute: String, filter-properties:Tuple<Key
                                         filter-properties]
   look-up-capacitors(attribute, params)
 
+; ========================================================
+; ==================== tantalum-cap =======================
+; ========================================================
+
+public defn tantalum-cap (user-params:Tuple<KeyValue>) -> Instantiable :
```
m-hilgendorf, line 183: You can add doc string for public functions to describe the API : 

```
doc: "Create a tantalum cap. Supported key/value args are \"capacitance\", \"tolerance\", etc."
public defn tantalum-cap (user-params:Tuple<KeyValue>) -> Instantiable :

doc: "Create a tantalum cap with given capacitance"
public defn tantalum-cap (cap:Double) -> Instantiable : 

doc: "Create a tantalum cap with a given capacitance and tolerance" 
public defn tantalum-cap (cap:Double, tolerance:Double) -> Instantiable : 
```

The `doc:` syntax is there for stanza's autodocumentation feature. We should start using it more so we can auto-generate OCDB's docs.

utils/generic-components.stanza
```diff
@@ -266,6 +176,41 @@ public defn look-up-ceramic-caps (attribute: String, filter-properties:Tuple<Key
...
+  val params = remove-duplicate-keys $ [["anode" => "tantalum" "min-rated-voltage" => MIN-CAP-VOLTAGE]
+                                        smd-query-properties()
+                                        user-params]
+  map(to-jitx, Capacitors(params, limit))
+
+; ========================================================
+; =============== look-up-tantalum-caps ===================
+; ========================================================
+
+public defn look-up-tantalum-caps (attribute: String) -> Tuple :
```
m-hilgendorf, line 205: What are the valid values for the `attribute` argument and what does it control? 

What is the return of this? Is it `Tuple<JSON>`?

utils/generic-components.stanza
```diff
@@ -266,6 +176,41 @@ public defn look-up-ceramic-caps (attribute: String, filter-properties:Tuple<Key
...
+                                        user-params]
+  to-jitx $ Capacitor(params)
+
+public defn tantalum-cap (cap:Double) -> Instantiable :
+  tantalum-cap(["capacitance" => cap])
+
+public defn tantalum-cap (cap:Double, tolerance:Double) -> Instantiable :
+  tantalum-cap(["capacitance" => cap "tolerance" => tolerance])
+
+public defn tantalum-caps (user-params:Tuple<KeyValue>, limit: Int) -> Tuple<Instantiable> :
```
m-hilgendorf, line 195: This function is interesting, it returns a tuple of valid components that satisfy your query. How would it be used in context, compared to using `lookup-tantalum-caps' and selecting one?

m-hilgendorf (commented):
We should have unit tests for these functions, something like this : 

```
#use-added-syntax(jitx, tests)
defpackage ocdb/test/tantalum-caps : 
  import core
  import collections
  import jitx
  import jitx/commands

deftest tantalum-cap-instantiation-1uf : 
  pcb-module module : 
     inst cap : tantalum-cap(1.0e-6)
     #EXPECT(property(cap.capacitance) == 1.0e-6)
     #EXPECT(property(cap.anode) == "tantalum")
   evaluate(module)
   
deftest tantalum-cap-lookup-1uf : 
  val caps = lookup-tantalum-cap(..., ["capacitance" => 1.0e-6])
  #ASSERT(not empty?(caps))
```

utils/db-parts.stanza
electricfungi: fixed

utils/generic-components.stanza
electricfungi: pass - this is a new doc flow beyond the scope of this change

utils/generic-components.stanza
electricfungi: pass - this gets captured in our docs

utils/generic-components.stanza
electricfungi: pass - this gets captured in our docs

utils/generic-components.stanza
m-hilgendorf: I don't see any docs for this.

m-hilgendorf (approved):

PR #185, opened 2021-12-03
added bundles, added trust level. Built Nov 2021

electricfungi:


components/tag-connect/TC2030-IDC.stanza
```diff
@@ -38,6 +38,7 @@ pcb-landpattern footprint :
 
 public pcb-component component :
   description = "Insertion point for TC-2030 connector"
+  trust = "built"
```
ariel-hi, line 41: I think you're looking for `property(self.trust) = "built"`

bhusang (approved):

PR #186, opened 2021-12-08
Docs/jitx 1032 examples to ocdb

electricfungi:
Add doc-examples.stanza, so folks can quickly visualize all the examples in the docs, and use ocdb to find examples instead of docs if they wish. Fixed issue with `gen-tant-cap-cmp` quick accessor input not matching `gen-cap-cmp`, added new `gen-cap-cmp` quick accessor.


m-hilgendorf (approved):
Looks good! Thanks for the comprehensive docs. 

I'm assuming the `ANCHOR` comments are used for the doc creation.

Only minor style nit is that functions with long call can be split into newlines, for example 

```stanza
   ind-strap(filter-5v, out-5v, ["inductance" => closest-std-val(20.0e-9,20.0) "tolerance" => 0.10 "type" => "Wirewound" "case" => get-valid-pkg-list("0603") "min-rated-temperature.max" => 40.0])
```

Can be

```
   ind-strap(filter-5v, out-5v, ["inductance" => closest-std-val(20.0e-9,20.0) 
                                            "tolerance" => 0.10 
                                            "type" => "Wirewound" 
                                            "case" => get-valid-pkg-list("0603") 
                                            "min-rated-temperature.max" => 40.0 ])
```

Which improves legibility. However I think we can leave that off to get this in.

PR #187, opened 2021-12-15
Updates to the ble-mote design and associated utilities

d-haldane:
Updated to a way of writing functional generators like `add-open-drain-pullups` that returns the components added to the design inside a module. Now you can access/group those components.

Also added an example of design rules to ble mote, updated ESP32 to not map input only pins to gpio, added some groups.

### DIFF
--- a/utils/generic-components.stanza
+++ b/utils/generic-components.stanza
@@ -966,10 +966,13 @@ public pcb-module micro-usb-connector :
   ; RC filter connection for chassis gnd
   res-strap(con.SHIELD, con.GND, 100.0e3)
   cap-strap(con.SHIELD, con.GND, 0.1e-6)
+  net SHIELD (con.SHIELD)
   net D+ (usb-2.data.P con.D+)
   net D- (usb-2.data.N con.D-)
   net ID (usb-2.id con.ID)
   property(con.VCC.power-supply-pin) = PowerSupplyPin(5.0, 0.5)
+
+  schematic-group(self) = micro-usb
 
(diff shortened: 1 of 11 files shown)


utils/generic-components.stanza
```diff
@@ -1194,15 +1211,22 @@ public pcb-module inverted-f-antenna :
...
+        val i = PWL(property(l.mcd-current))[brightness]
+        closest-std-val((vio - property(l.forward-voltage)) / i, 5.0)
+      public inst led : ocdb/foshan-optoelectronics/FM-B2020RGBA-HG/component ; TODO: make generic
+      
+      res-strap(pwm[0].timer, led.l.r, calc-r(led.l.r, brightness, vio))
+      res-strap(pwm[1].timer, led.l.g, calc-r(led.l.g, brightness, vio))
+      res-strap(pwm[2].timer, led.l.b, calc-r(led.l.b, brightness, vio))
+
+    inst rgb:rgb-indicator
     require pwm:timer[3] from proc
```
bhusang, line 1228: You have a port pwm : timer[3] and then you overload pwm with a require statement? I am confused about which one gets connected to the rgb.

d-haldane: Good point - I've changed the names to avoid conflating timer and pwm

A module is defined with a port `control-pins` of type `timer[3]`. Later I use a require statement to pull a port of that type of the processor and connect it to the module port. So in the end, three pins supporting timer are attached to the led.

bhusang (approved):
Looks good, with nice examples to copy

PR #193, opened 2022-01-08
Adding pin props and script for ioc file generation (JITX-937)

bhusang:
This is a work in progress but I would like to get some early feedback. My code in the generate-ioc.stanza is brute-force coding for the most part but I am a hardware guy :). I will likely add a demo script at some point based off the design that I have done.


tests/test-stm32-mcu-props.stanza
```diff
@@ -0,0 +1,164 @@
...
+  val lf-cb = add-xtal-caps(xtal, gnd, 0.5e-12)
+  
+  net (lf-osc.in xtal.p[1])
+  net (lf-osc.out xtal.p[2])
+
+  net (debug mcu-debug)
+
+  net (mcu-debug.swdio mcu-debug.swdclk)
+
+  net dummy ()
```
ariel-hi, line 105: We have an underused JITX concept called `node` that just names something you can net to.
```
node dummy : pin
```

ariel-hi: Of course a dummy net works too.

tests/test-stm32-mcu-props.stanza
```diff
@@ -0,0 +1,164 @@
...
+
+  require mcu-gpio-o:gpio[16] from mcu
+  for i in 0 to 16 do :
+    net (mcu-gpio-o[i].gpio dummy)
+    property(mcu-gpio-o[i].gpio.stm32-gpio-iomode) = ocdb/st-microelectronics/generate-ioc/GPIOOutput
```
ariel-hi, line 110: I believe you should be able to write `GPIOOutput` instead of the fully qualified package name. (Same for the other loops below)

components/st-microelectronics/STM32F407ZET6.stanza
```diff
@@ -0,0 +1,534 @@
...
+  supports SPI2_MISO:
+    SPI2_MISO.p => self.PC[2]
+  supports SPI2_MISO:
+    SPI2_MISO.p => self.PB[14]
+
+  supports spi-controller():
+    require sdo-pin: SPI2_MOSI
+    require sdi-pin: SPI2_MISO
+    require sck-pin: SPI2_SCK
+    require cs-pin: SPI2_NSS ; how to make this optional?
```
ariel-hi, line 517: The `spi-controller` has a `cs` pin. If this require were optional, what pin would be used for `cs`?

components/st-microelectronics/generate-ioc.stanza
```diff
@@ -0,0 +1,351 @@
...
+    USARTSmartcardWClk :  "SmartCard_With_Clock"
+    USARTLIN :            "LIN"
+
+defn stm32-gpio-mode (mode : GPIOModes) -> String :
+  switch(mode) :
+    GPIOOutput :   "GPIO_Output"
+    GPIOInput :    "GPIO_Input"
+    GPIOAnalog :   "GPIO_Analog"
+    GPIOEventOut : "EVENTOUT"
+    GPIOExtI :     "GPXTI\n# can't currently write EXTI pins correctly."
```
ariel-hi, line 156: What's the limitation?

components/st-microelectronics/generate-ioc.stanza
```diff
@@ -0,0 +1,351 @@
...
+
+defn compose-ioc-contents (i : JITXObject) :
+  val ip-template = "Mcu.IP%_=%_\n"
+  val pin-template = "Mcu.Pin%_=%_\n"
+  var content : String = ""
+  var pin-content : String = ""
+  var pin-params : String = ""
+  var ip-content : String = ""
+  val ip-names = HashSet<STM32IPs>()
+  var exist-params = false
```
ariel-hi, line 179: I suggest naming this `exist-params?` to match Standard Stanza for boolean variable names.

components/st-microelectronics/generate-ioc.stanza
```diff
@@ -0,0 +1,351 @@
...
+              if has-property?(p.stm32-gpio-iomode) :
+                pin-content = append(pin-content, to-string("%_.Signal=%_\n" % [property(p.stm32-pin-name), stm32-gpio-mode(property(p.stm32-gpio-iomode))]))
+              else if has-property?(p.stm32-signal-name) :
+                pin-content = append(pin-content, to-string("%_.Signal=%_\n" % [property(p.stm32-pin-name), property(p.stm32-signal-name)]))
+                ; special case of signals with names and also used as ADC
+                if has-property?(p.stm32-adc-mode) :
+                  pin-content = append(pin-content, to-string("%_.Mode=%_\n" % [property(p.stm32-pin-name), property(p.stm32-adc-mode)]))
+              if has-property?(p.stm32-gpio-speed) :
+                if exist-params == true :
+                  pin-params = append(pin-params, to-string(","))
```
ariel-hi, line 234: I like how you're checking `has-property?`.

Small details:
- Instead of `if boolean == true`, you can write `if boolean`
- Strings like "," and "GPIO_PuPd" without string formats (%_) don't need `to-string`. They're already Strings.

components/st-microelectronics/generate-ioc.stanza
```diff
@@ -0,0 +1,351 @@
...
+                  LSEExternalMode :
+                    pin-content = append(pin-content, to-string("%_.Mode=LSE-External-Oscillator\n" % [property(p.stm32-pin-name)]))
+              if has-property?(p.stm32-pin-locked) :
+                if property(p.stm32-pin-locked) :
+                  pin-content = append(pin-content, to-string("%_.Locked=true\n" % [property(p.stm32-pin-name)]))
+
+              ; output data for a pin now
+              if exist-params == true :
+                content = append(content, to-string("%_\n" % [pin-params]))
+              if length(pin-content) > 0 :
```
ariel-hi, line 280: `if not empty?(pin-content)`

components/st-microelectronics/generate-ioc.stanza
```diff
@@ -0,0 +1,351 @@
...
+  import ocdb/checks
+  import ocdb/generic-components
+
+  import jitx/powergen/gen
+
+  import ocdb/design-vars
+  import ocdb/property-structs
+  import ocdb/bundles
+
+public pcb-enum ocdb/st-microelectronics/generate-ioc/GPIOModes :
```
d-haldane, line 17: Since this is an enum and the elements are global, you may want to prefix with STM e.g. `STMGPIOOutput`

Or if you feel these should be used for everything - move the enum to somewhere common like property-structs.

ariel-hi (changes requested):
Nice! I've written some comments.

components/st-microelectronics/STM32F407ZET6.stanza
```diff
@@ -0,0 +1,534 @@
...
+
+  for i in ADC3-UNIQUE-LIST do :
+    val p = ADC3-LIST[i]
+    supports adc:
+      adc.adc => p
+      property(p.stm32-signal-name) = to-string("ADC3_IN%_" % [i])
+      property(p.stm32-adc-mode) = to-string("IN%_" % [i])
+      property(p.stm32-ip-name) = ocdb/st-microelectronics/generate-ioc/ADC3IP
+
+  for i in 0 to 16 do :
```
d-haldane, line 250: Sometimes ports like `PA` can be sparse (e.g. missing `PA[8]` and `PA[14]`) so you could use a code pattern like

```
for p in pins(self.PE) do :
  supports gpio :
    gpio.gpio => p
```

and save some difficulty.

components/st-microelectronics/STM32F407ZET6.stanza
d-haldane: I think we covered this already. The cs pin is optional and you would avoid it by requesting spi[ pins I want])

tests/test-stm32-mcu-props.stanza
bhusang: I just use the dummy net so that the pins remain connected. Would I use your node definition above as a net in the net statement?

node dummy : pin

net (dummy mcu.somepin)

components/st-microelectronics/generate-ioc.stanza
bhusang: There is a number associated with the GPIOExt statement that I believe we do not know a-priori.

components/st-microelectronics/generate-ioc.stanza
bhusang: Done.

components/st-microelectronics/generate-ioc.stanza
bhusang: Danke schon.

components/st-microelectronics/generate-ioc.stanza
bhusang: For now I think that I will prefix the enums as I don't know if these names will make sense in the global sense (yet).

tests/test-stm32-mcu-props.stanza
bhusang: Good suggestion -- improves readability as well. Done.

components/st-microelectronics/STM32F407ZET6.stanza
bhusang: Ok, nice suggestion. I will incorporate it.

bhusang: Hmm.. I seem to have run into a problem with the pin property assignment:

  for p in pins(self.PA) do :
    supports gpio :
      gpio.gpio => p
      property(self.p.stm32-signal-name) = "GPIO"

complains as follows:
Uncaught Exception: Cannot access named field 'p'.
  in ocdb/st-microelectronics/STM32F407ZET6
    at ../open-components-database/components/st-microelectronics/STM32F407ZET6.stanza:253.15
  in ocdb/st-microelectronics/STM32F407ZET6
    at ../open-components-database/components/st-microelectronics/STM32F407ZET6.stanza:252.6
  in ocdb/st-microelectronics/STM32F407ZET6
    at ../open-components-database/components/st-microelectronics/STM32F407ZET6.stanza:250.25
  in ocdb/tests/test-stm32-mcu-props
    at test-stm32-mcu-props.stanza:40.7
  in ocdb/generator-utils/run-final-passes
    at ../open-components-database/utils/generator-utils.stanza:160.28
  in ocdb/tests/test-stm32-mcu-props
    at test-stm32-mcu-props.stanza:153.19

d-haldane:
For setting the property, you want property(p....

Not property(self.p....

components/st-microelectronics/STM32F407ZET6.stanza
bhusang: Oh duh. Thanks for the fix.

tests/test-stm32-mcu-props.stanza
ariel-hi: Yes, it would work just like that!

ariel-hi (approved):
Thanks for considering my feedback!
It's awesome that you've exported your STM JITX project to .ioc

PR #201, opened 2022-03-03
Only box some units, use bank property

ariel-hi:
`assign-symbols` is a new utility in `ocdb/box-symbol`.

Use this inside a `pcb-component` to assign each bank to a `pcb-symbol` or to a box symbol.

Example:
```
  assign-symbols([
    Ref("A") => ECP5U25-MG285A
    Ref("B") => ECP5U25-MG285B
    Ref("C") => ECP5U25-MG285C
    Ref("D") => ECP5U25-MG285D
    Ref("E") => ECP5U25-MG285E
    Ref("F") => ECP5U25-MG285F
    Ref("G") => false
    Ref("H") => ECP5U25-MG285H
    Ref("I") => ECP5U25-MG285I])
```
assigns banks A through I to `pcb-symbols` where all components with that bank property are mapped to a symbol pin with that name.
It is an error if the component pin names do not match the symbol pin names.
Note that G is false, meaning that there is no corresponding `pcb-symbol`, instead a box symbol is created.

### DIFF
--- a/utils/box-symbol.stanza
+++ b/utils/box-symbol.stanza
@@ -307,14 +307,17 @@ public defn assign-symbols (banks:Tuple<KeyValue<Int|Ref, UnitSymbol>>) :
           (box:BoxSymbol) : None()
           (symbol:SchematicSymbol) :
             val symbol-pins = to-hashset(ref, /pins(symbol))
-            val mismatched-pins =
+            val mismatched-pins = to-tuple $
               for (p in pins, b in pin-banks) seq? :
                 val pin-ref = ref(p)
                 if value!(b) == bank and not symbol-pins[pin-ref] :
                   One(pin-ref)
                 else :
                   None()
-            One(name(symbol) => to-tuple(mismatched-pins))
+            if not empty?(mismatched-pins) :
+              One(name(symbol) => to-tuple(mismatched-pins))
+            else :
+              None()
     if not empty?(bad-matches) :
       throw(MissingSymbolPins(bad-matches))
   ensure-non-empty-banks!()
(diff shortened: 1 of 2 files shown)


utils/box-symbol.stanza
```diff
@@ -244,70 +253,14 @@ public defn make-box-symbol (banks:Seqable<BoxBank>) :
...
-  ;Sort the given pins by their .pin-properties-row-index field.
-  defn sort-by-row-index (pins:Tuple<JITXObject>) -> Tuple<JITXObject> :
-    ;Compare pin[i0] versus pin[i1].
-    ;Indices are used, instead of the pin directly, in order to perform a
-    ;stable sort.
-    ;All pins without a .pin-properties-row-index are put in the beginning.
-    defn compare-index (i0:Int, i1:Int) :
-      val ai = value?(property?(pins[i0].pin-properties-row-index))
-      val bi = value?(property?(pins[i1].pin-properties-row-index))
-      match(ai, bi) :
```
ariel-hi, line 256: These functions are shared by `assign-symbols`, so the helper functions / checkers are moved to ~ line 346.

PhilippeFerreiraDeSousa:
The change from
```
public defn pins (def:Instantiable|Bundle) :
  pins(JITX-CLIENT, def)
```
to
```
public defn pins (def:Instantiable|Bundle|SchematicSymbol) :
  pins(JITX-CLIENT, def)
```
needed in jitx/commands is on the importer branch but not on develop

PhilippeFerreiraDeSousa (approved):
Great!

PR #202, opened 2022-03-03
added NVIC setup to generation ioc that was missing

bhusang:


components/st-microelectronics/generate-ioc.stanza
```diff
@@ -87,6 +87,10 @@ defn ioc-file-header () -> String :
   val header = "#MicroXplorer Configuration settings - do not modify\nFile.Version=6\nboard=custom\n"
   header
 
+defn default-nvic-settings () -> String :
+  val body = "NVIC.ForceEnableDMAVector=true\nNVIC.HardFault_IRQn=true\\:0\\:0\\:false\\:false\\:true\\:false\\:false\nNVIC.NonMaskableInt_IRQn=true\\:0\\:0\\:false\\:false\\:true\\:false\\:false\nNVIC.PendSV_IRQn=true\\:0\\:0\\:false\\:false\\:true\\:false\\:false\nNVIC.SVC_IRQn=true\\:0\\:0\\:false\\:false\\:true\\:false\\:false\nNVIC.SysTick_IRQn=true\\:0\\:0\\:false\\:false\\:true\\:false\\:true\n"
```
ariel-hi, line 91: You can use a string block to split the body across multiple lines and to not escape backslash:
```
val body = \<nvic-body>NVIC.ForceEnableDMAVector=true
NVIC.HardFault_IRQn=true\:0\:0\:false\:false\:false\:true\:false
NVIC.NonMaskableInt_IRQn=true\:0\:0\:false\:false\:false\:true\:false
NVIC.PendSV_IRQn=true\:0\:0\:false\:false\:true\:true\:false
NVIC.SVCall_IRQn=true\:0\:0\:false\:false\:true\:true\:false
NVIC.SysTick_IRQn=true\:0\:0\:false\:false\:true\:true\:true
<nvic-body>
```

ariel-hi: Oops, don't use my exact text. I've copied some of the wrong values. The idea remains.

PhilippeFerreiraDeSousa: In a function, you would just do
```
defn default-nvic-settings () -> String :
  \<nvic-body>NVIC.ForceEnableDMAVector=true
NVIC.HardFault_IRQn=true\:0\:0\:false\:false\:true\:false\:false
NVIC.NonMaskableInt_IRQn=true\:0\:0\:false\:false\:true\:false\:false
NVIC.PendSV_IRQn=true\:0\:0\:false\:false\:true\:false\:false
NVIC.SVC_IRQn=true\:0\:0\:false\:false\:true\:false\:false
NVIC.SysTick_IRQn=true\:0\:0\:false\:false\:true\:false\:true
<nvic-body>
```
But you may not need to create a function to return a constant, this is uncommon

ariel-hi (approved):
Great! Thanks for follow-up changes!

PR #206, opened 2022-04-04
Feature/jitx 1462 artwork

electricfungi:
Add a framework for mechanical BOM items such as labels and hardware, and for artwork. Add example artwork


d-haldane:
Can you store geometry separately from the land patterns? 

e.g. in esd-symbol you have the same geometry copied twice for a silkscreen and a copper version. Better would be to store geometry as a Union, and then pass an argument for the layer to put the geometry on.

This pattern is repeated a few times, and the duplicate info is blowing up your code size.

Also, try to use better arguments. e.g. instead of `"copper"` as an argument, pass the actual layer. If that's not always possible, at least use an `enum` instead of a string.

electricfungi:
Passing arbitrary layers inside each artwork and drawing lines of min-width for the layer. This requires jitx-client PR#1004 to run

electricfungi:
PR has been merged this should work with latest dev

utils/mechanical.stanza
```diff
@@ -0,0 +1,268 @@
...
+  layer("foreground") = Line(1.0, [Point(1.89669, 2.46568), Point(2.96349, 2.87884)])
+  layer("foreground") = Line(1.0, [Point(1.89669, 2.967), Point(2.96349, 3.38015)])
+  layer("foreground") = Line(1.0, [Point(1.99829, 3.04274), Point(2.86189, 3.41253)])
+  layer("foreground") = Line(1.0, [Point(1.99829, 2.93462), Point(2.86189, 3.30441)])
+  layer("foreground") = Line(1.0, [Point(1.99829, 3.43594), Point(2.86189, 3.80573)])
+  layer("foreground") = Line(1.0, [Point(1.99829, 3.54406), Point(2.86189, 3.91385)])
+  layer("foreground") = Line(1.0, [Point(1.89669, 3.46832), Point(2.96349, 3.88147)])
+  layer("foreground") = Line(1.0, [Point(1.89669, 1.95768), Point(2.96349, 2.37084)])
+  layer("foreground") = Line(1.0, [Point(1.99829, 2.03342), Point(2.86189, 2.40322)])
+  layer("foreground") = Line(1.0, [Point(1.99829, 1.9253), Point(2.86189, 2.2951)])
```
PhilippeFerreiraDeSousa, line 110: How did you generate all this code? Did you use python or something? Same for artworks

electricfungi: These are imported geometries

components/artwork/esd-symbol.stanza
```diff
@@ -0,0 +1,2764 @@
...
+  import jitx
+  import jitx/commands
+  import ocdb/defaults
+  import ocdb/land-patterns
+  import ocdb/generic-components
+  import ocdb/symbols
+
+pcb-landpattern esd-lp (lyr:LayerSpecifier|LayerIndex, side:Side|False) :
+  if side is Side :
+    layer(Courtyard(side as Side)) = Rectangle(23.368, 20.9296, loc(-2.2098, 2.6416))
```
ariel-hi, line 24: Another option in Stanza:
If you have a variable (`side`) that you'd like to check its type (`side is Side`) and cast its type (`side as Side`), you can write:
```
match(side:Side) :
  layer(Courtyard(side)) = Rectangle(23.368, 20.9296, loc(-2.2098, 2.6416))
```

This is already good, and it's not necessary to change or even to follow this suggestion in the future.

components/artwork/esd-symbol.stanza
```diff
@@ -0,0 +1,2764 @@
...
+  val sym = ESD-symbol(show-symbol)
+  symbol = sym()
+  reference-prefix = "ESD"
+
+; Defaults
+public defn regular () :
+  regular(Silkscreen("artwork", Top),true, Top)
+public defn regular (layer:LayerSpecifier|LayerIndex) :
+  regular(layer,true, Top)
+public defn regular (layer:LayerSpecifier|LayerIndex,show-symbol:True|False) :
```
ariel-hi, line 52: I would have thought that these entrypoints should be more descriptive, like `regular-esd-symbol()`.
Now that I see their usage in `ble-mote`, it reads better like this.
These should be always accessed with qualified packages `ocdb/esd-symbol/regular()` because of common name collisions with other packages.

i.e. The following requires prefixing:
```
import ocdb/esd-symbol
; --> Use prefix ---- with: prefix(large) => esd-symbol-
import ocdb/jitx-logo
; --> Use prefix ---- with: prefix(large) => jitx-logo-

esd-symbol-large()
```

Is it the expected workflow for a designer to write `ocdb/esd-symbol/regular()` instead of importing `ocdb/esd-symbol`?

ariel-hi (commented):
Great symbols, and I like your parameterize land patterns!

components/artwork/esd-symbol.stanza
electricfungi: Exactly, typically, components in ocdb are named "component". here I named them small / large to describe size. It's not expected to add a dedicated import statement for any ocdb component, but if someone did on these or most components in ocdb, they would likely need a prefix to avoid collisions with other components. General workflow is for designers to use fully qualified names:
`  inst pcb-name: ocdb/board-text/text("PCB, BLE MOTE", 1.5, clearance(current-rules(), MinSilkscreenWidth) )
  inst logo : ocdb/jitx-logo/small(Silkscreen("artwork", Top),true)
  inst assy-label : ocdb/mechanical/gen-label()`

components/artwork/esd-symbol.stanza
```diff
@@ -0,0 +1,6189 @@
...
+defpackage ocdb/esd-symbol :
+
+; Symbols: regular, large
+; Options to pass: (layer:String,show-symbol:True|False)
+; How to use:
+; create regular esd symbol on copper layer, show ESD symbol in schematic:
+;   inst <name> : ocdb/esd-symbol/regular("copper",true) 
+; create large esd symbol on silkscreen layer, show box symbol in schematic:
+;   inst <name> : ocdb/esd-symbol/large ("silkscreen",false) 
+; use place function to control position, orientation, and top/bottom layer
```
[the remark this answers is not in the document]
electricfungi: Fixed. Also - I noticed some inconsistencies in documentation and order of code and made it more consistent

components/artwork/esd-symbol.stanza
electricfungi: noted

ariel-hi (approved):
Great, thanks for the consistent documentation!

PR #207, opened 2022-04-04
Bg/improve tolerance mnemonics

bhusang:
Improved the tolerance tol%() function to take as arguments a percentage 0 <= x <= 100 and give a warning when the argument is less than 1.0 (which is still valid)

Also changed typ() to typ-value() to match min-value() and max-value()


modules/passive-circuits.stanza
```diff
@@ -124,9 +124,9 @@ defn make-voltage-divider-module (in: JITXObject, out: JITXObject, lo: JITXObjec
     net (r-lo.p[1] r-hi.p[2] out)
     net (r-lo.p[2] lo)
 
-    val vin = typ(vo) * (property(r-lo.resistance) + property(r-hi.resistance)) / property(r-lo.resistance)
-    property(r-lo.operating-point) = OperatingPoint(Interval(0.0, typ(vo), false),       Interval(0.0, typ(vo) / property(r-lo.resistance), false))
-    property(r-hi.operating-point) = OperatingPoint(Interval(0.0, vin - typ(vo), false), Interval(0.0, (vin - typ(vo))/ property(r-hi.resistance), false))
+    val vin = typ-value(vo) * (property(r-lo.resistance) + property(r-hi.resistance)) / property(r-lo.resistance)
+    property(r-lo.operating-point) = OperatingPoint(Interval(0.0, typ-value(vo), false),       Interval(0.0, typ-value(vo) / property(r-lo.resistance), false))
```
electricfungi, line 128: I don't think we're supporting interval anymore

bhusang: You are right -- that's for the next PR.

utils/tolerance.stanza
```diff
@@ -3,36 +3,56 @@ defpackage ocdb/tolerance:
   import core
   import collections
   import math
+  import jitx/utils/logging
```
m-hilgendorf, line 6: This API is an undocumented internal library used within the JITX compiler and not public facing, it may change or be removed in the future. I think we should try and refrain from using these where possible.

utils/tolerance.stanza
```diff
@@ -3,36 +3,56 @@ defpackage ocdb/tolerance:
...
+; Create a symmetric tolerance from percentage 
+; where the percentage ranges from 0.0 to 100.0 (not 0.0 to 1.0)
 public defn tol% (typ:Double, tol:Double) -> Toleranced:
-  val check = min-typ-max(0.0, 0.5, 1.0)
-  fatal("tol must be in range 0.0 <= tol <= 1.0") when not in-range?(check, tol)
-  Toleranced(typ, typ * tol, typ * (-1.0 * tol))
+  val check100 = min-typ-max(0.0, false, 100.0)
+  val check1 = min-typ-max(0.0, false, 1.0)
+  fatal("tol must be in range 0.0 <= tol <= 100.0") when not in-range?(check100, tol)
+  warning("tol%(typ, tol) -> tol might be mis-specified (< 1.0), expecting 0.0 <= tol <= 100.0") when in-range?(check1, tol)
```
m-hilgendorf, line 52: see the comment above, I don't think we need a warning here.

m-hilgendorf (changes requested):

utils/tolerance.stanza
bhusang: Ok, I understand, but I need a method to inform the user that a parameter that they have entered *might* be too small for the calculation that they are doing. This is important for user comprehension. Putting a fatal() in here would also *not* be a good choice. Print/println would also seem to be wrong because we want to be able at some point to log all warnings in a report.

utils/tolerance.stanza
bhusang: We absolutely need a warning to the user -- if they use 0.1 as the argument, did they mean 10% or 0.1%? More often it will be 10% but we need to hold our user's hands.

utils/tolerance.stanza
bhusang: @m-hilgendorf Should I change it to println() for now? Or leave it as a warning() with a note that it will be improved?

@d-haldane I have used warning() to provide user feedback but still allow the compilation to continue. Sounds like we need a deeper look at the user feedback system with differentiated levels (fatal, error, warning, info). Ultimately these need to be logged somewhere so that an audit report can be created with all of the compilation information present. Just capturing stdout/stderr seems deficient for the future.

d-haldane: `println` for now to get this through.

It's a good point that we need a user logging system. How do we replace the little notes people like to leave on schematics? How do we keep track of assumptions?

m-hilgendorf: The advantage to using the range [0,1] is that it is unambiguous. >1 is always incorrect in that sense. 

I agree with @d-haldane that println is appropriate here until we have a robust, general purpose logging library. We need it in any case.

m-hilgendorf: You can also println to current-error-stream()

utils/tolerance.stanza
m-hilgendorf: So why not leave the original implementation?

utils/tolerance.stanza
bhusang: I think the question of [0 1] versus [0 100] can be resolved by thinking as a user when I see a function called tol%() what would I think that the argument would be? It does *not* imply right off the bat that it needs to be less than 1.0. That's the issue -> if a user tries to use the function, how will they know that the argument is in the range [0.0 1.0]? I prefer (as a user) to have an argument range that "semantically" matches the function name.

m-hilgendorf: Very good point. After a bit of thinking the alternative is still ambiguous, just in the opposite direction.

bhusang: Agreed.

m-hilgendorf (approved):
Thank you, @bhusang !

PR #213, opened 2022-04-21
Fixing toleranced percentages as well as improving the tutorial code

bhusang:


components/unisonic/lm317a.stanza
```diff
@@ -42,3 +44,6 @@ public pcb-component component :
   property(self.reference-voltage) = min-typ-max(1.20, 1.25, 1.3)
   property(self.minimum-load) = 4.5e-3
   property(self.adj-current) = 50.0e-6
+  property(self.max-current) = 1.5 ; A
+  ; property(self.output.power-pin) = PowerSupplyPin(v-out, 0.5)
```
d-haldane, line 48: Do these commented out properties need to be in the component model?

designs/tutorial.stanza
```diff
@@ -14,19 +14,31 @@ defpackage ocdb/designs/tutorial :
 
 val board-shape = RoundedRectangle(35.0, 30.0, 0.5)
 
+; this is a simple usage of the LM317A LDO where the component is instantiated with a simple
+; resistor divider feedback 
 pcb-module simple-lm317a-regulator (v-out:Double) :
+  port pwr-in : power
```
d-haldane, line 20: For the very first, 5 minute version of the tutorial I started without ports.

There will be a sequence of tutorials building up to the full test fixture design, and I was going to cover ports in the one immediately following. What do you think about that?

It is weird to have a module with no ports, but I was trying to minimize the overall number of concepts delivered in 5 minutes.

components/unisonic/lm317a.stanza
bhusang: I left the commented out because I was unsure how to set them correctly. At the component level we don't know the target output voltage. If they become useless, I can delete the properties

designs/tutorial.stanza
bhusang: Yea, this is a good question -- I think that ports are useful because if someone copies the simple regulator as a template for their design, it will not be useful without ports. My mindset was that the tutorial code will be copied and extended by the user because it is a convenient jumping off point. Hence the inclusion of ports.

components/unisonic/lm317a.stanza
d-haldane: Let's delete for now - Mike has pointed out commented code has a lot of ambiguities, and can be added back pretty easily

bhusang: Ok, np.

designs/tutorial.stanza
d-haldane: Let's focus this design on the quick start tutorial. 

I think that design templates are *super* important, but we do have a better way to create new projects with the new extension features.

There will be a version with ports in the same file no matter what. It does sound like we can avoid ports in the simple regulator.

bhusang: Ok, np.

bhusang:
@d-haldane Please take a look when convenient -- should be ready to go. Tutorial runs and the tolerance checks pass.

d-haldane (approved):

PR #214, opened 2022-04-27
Feature/jitx 1573 smd pad

electricfungi:
Expanded SMD pad and PTH pad to accept arbitrary shapes for copper, solder mask, and paste layer. In the case of pth pad, solder mask layers are mirrored on top and bottom, copper layers are on all layers for pth.

### DIFF
--- a/utils/land-patterns.stanza
+++ b/utils/land-patterns.stanza
@@ -270,12 +270,22 @@ public pcb-pad testpoint-pad (testpoint-shape:Shape):
 public defn testpoint-pad (diameter:Double) :
   testpoint-pad(Circle(0.5 * diameter))
 
-public pcb-pad pth-pad (pad-shape:Shape, hole-shape:Shape):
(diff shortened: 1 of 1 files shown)


utils/land-patterns.stanza
```diff
@@ -217,12 +217,19 @@ public defn soldermask-defined (lp:LandPattern):
 
   SMD-l
 
-public pcb-pad smd-pad (s:Shape):
+public pcb-pad smd-pad (copper-layer:Shape,solder-mask-layer:Shape|False,paste-layer:Shape):
```
m-hilgendorf, line 220: I like it! 

We also should update the docs, but that is outside this repo so out of scope for this PR.

m-hilgendorf (approved):
Great! I like the additional customization.

utils/land-patterns.stanza
electricfungi: Created ticket to capture this

PR #216, opened 2022-04-28
Edits to Checks and Property Struct Printers

ariel-hi:
This PR reflects the growing progress of the new #CHECK macro.
Added printers to some property structs.


utils/property-structs.stanza
```diff
@@ -136,4 +136,16 @@ public pcb-struct ocdb/property-structs/CrystalResonator :
...
+; ================================
+;[TODO] Write print methods for other property structs.
+defmethod print (o:OutputStream, i:Interval) :
+  print{o, _} $
+    match(nominal(i)) :
+      (nominal:Double) : "[%_, (%_), %_]" % [min(i), nominal, max(i)]
+      (f:False) : "[%_, %_]" % [min(i), max(i)]
+
+defmethod print (o:OutputStream, r:RatedTemperature) :
+  print(o, operating-temperature(r))
```
ariel-hi, line 151: Alternate printers are welcome!
```
Interval(1.0, 2.0)
> [1.0, 2.0]

Interval(1.0, 1.5, 2.0)
> [1.0, (1.5), 2.0]

RatedTemperature(Interval(5.0, 7.0))
> [5.0, 7.0]
```

utils/checks.stanza
```diff
@@ -64,8 +67,8 @@ public pcb-check within? (t:Toleranced, value:Toleranced) :
     subcheck-description = "Check that a value is within a given range"
     pass-message =         "Value %_ to %_ is within the range of %_ to %_" % [min-value(t) max-value(t) min-value(value) max-value(value)]
     fail-message =         "Value %_ to %_ is not within the range of %_ to %_" % [min-value(t) max-value(t) min-value(value) max-value(value)]
-    locators =             [t value]
     )
+    ; locators =             [t value]
```
ariel-hi, line 71: Locators aren't yet supported in the new macro, so I have commented them out in the meantime!

utils/checks.stanza
```diff
@@ -94,25 +97,25 @@ public pcb-check rated-temperature (component:JITXObject):
   val temp = operating-temperature(property(component.rated-temperature))
   ; println("Rating %_" % [temp])
   #CHECK(
-    condition =            in-range?(temp, OPERATING-TEMPERATURE[0]),
+    condition =            contains?(temp, OPERATING-TEMPERATURE[0]),
```
ariel-hi, line 100:
```
public defn contains? (a:Interval, b:Double) -> True|False :
  min(a) <= b and
  max(a) >= b
```

utils/checks.stanza
```diff
@@ -167,7 +170,7 @@ public pcb-check io (module:JITXObject) :
               ; println("checking %_ vs %_" % [ref(out-p) ref(in-p)])
               check-single-push-pull-levels(out-p, in-p)
 
-        if o is OpenCollector :
+        match(o:OpenCollector) :
```
ariel-hi, line 173: Using a `match` statement to cast `o` to an `OpenCollector`.

utils/checks.stanza
```diff
@@ -408,23 +412,31 @@ public defn check-single-push-pull-levels (out-p:JITXObject, in-p:JITXObject) :
...
   val description = "Check that the object(s) are connected"
   val category = "net-check"
+  val pass-message =
+    match(p) :
+      (p:JITXObject) : "Object %_ is connected" % [ref(p)]
+      (p:Tuple<JITXObject>) : "Objects %, are connected" % [map(ref, p)]
+  val fail-message =
+    match(p) :
+      (p:JITXObject) : "Object %_ is not connected" % [ref(p)]
+      (p:Tuple<JITXObject>) : "Objects %, are not connected" % [map(ref, p)]
```
ariel-hi, line 429: We probably want the messages to change depending on what's passed in!
We could also branch based on the number of elements in the tuple. Zero, one, or multiple could be handled differently.

utils/checks.stanza
```diff
@@ -1166,7 +1178,7 @@ public pcb-check check-min-copper-copper-space (landpattern:LandPattern, side:Si
   val description = "Check min copper spacing"
   val category = "landpattern-check"
 
-  val copper = to-seq(copper(landpattern, side))
+  val copper = to-tuple(copper(landpattern, side))
```
ariel-hi, line 1181: `to-seq` converts a `Seqable` to a `Seq`.
These objects are streams, and their elements are consumed / disappear after being read once.
The #CHECK condition consumes the stream by calling `min-space(copper)`.
The pass/fail messages error because the `copper` shapes are now empty.

Some options:
- Use `to-tuple` instead of `to-seq`. A Tuple is a Seqable!
- Pull the `min-space` calculation out and store in a variable.

utils/checks.stanza
```diff
@@ -1270,7 +1282,7 @@ public defn check-landpattern (component:JITXObject|JITXDef) :
     for side in [Top, Bottom] do :
       check check-min-copper-copper-space(lp, side)
       check check-min-copper-width(lp, side)
-      check check-min-solder-mask-sliver(layer(lp, SolderMask(side)))
+      check check-min-solder-mask-sliver(lp, layer(lp, SolderMask(side)))
```
ariel-hi, line 1285: Passing the land pattern into `check-min-solder-mask-sliver` so its pass/fail messages have access to it.

utils/checks.stanza
bhusang: Good idea!

utils/checks.stanza
bhusang: Oh, great catch, this issue has been around for a while (I did not edit this).

utils/property-structs.stanza
bhusang: There is a printer in tolerance.stanza for Toleranced() objects. Not sure that we need it here?

utils/checks.stanza
bhusang: Gotcha

utils/checks.stanza
bhusang: Deprecated!

utils/checks.stanza
bhusang: Nice!

utils/checks.stanza
bhusang: Awesome!

bhusang:
I would update your version of my branch. I *think* that I have included all of your changes there.

utils/property-structs.stanza
ariel-hi: Right, we won't need another printer for `Toleranced`.
We should still have one for `RatedTemperature`.

bhusang: Oh, I see. RatedTemperature is a type Toleranced now.

ariel-hi:
Thanks @bhusang I'll grab the updates from your branch!

utils/checks.stanza
```diff
@@ -706,23 +720,24 @@ public pcb-check resistor-component (r:JITXObject) :
...
-      ; locators =             [r]
 
 public pcb-check capacitor-component (c:JITXObject) :
   val name = "Capacitor checks"
   val description = "Check capacitor component properties"
   val category = "component-check"
+  println("HERE")
   #CHECK(
-    condition =            has-property?(c.capacitor),
+    condition =            (println("COND") has-property?(c.capacitor)),
```
bhusang, line 732: What is the purpose of the println()?

utils/property-structs.stanza
bhusang: So all we need now is:
```
defmethod print (o:OutputStream, r:RatedTemperature) : 
  print(o, operating-temperature(r))
```

bhusang (changes requested):
@ariel-hi Looks great (I really like your changes in checks.stanza to simplify the code) -- one remaining question in checks.stanza about the println() and the need to add the print for RatedTemperature.

utils/property-structs.stanza
ariel-hi: Added!

ariel-hi:
@bhusang Updated! The println got in by mistake as I was debugging.

utils/checks.stanza
```diff
@@ -64,14 +64,13 @@ public pcb-check within? (t:Toleranced, value:Toleranced) :
     subcheck-description = "Check that a value is within a given range"
     pass-message =         "Value %_ to %_ is within the range of %_ to %_" % [min-value(t) max-value(t) min-value(value) max-value(value)]
     fail-message =         "Value %_ to %_ is not within the range of %_ to %_" % [min-value(t) max-value(t) min-value(value) max-value(value)]
-    ; locators =             [t value]
+    locators =             []
```
ariel-hi, line 67: Locators can be `JITXObject` or `JITXDef`. Locators do not support for `pcb-structs`.
Let me know if that diverges from MVP.

bhusang (approved):

PR #218, opened 2022-05-02
Refactored `pcb-check io` to split up and updated some code to be mor…

m-hilgendorf:
…e idiomatic

- Added `check ... when has-property` syntax to `check-design`
- Added doc string for `check-design` with current expected behavior
- Changed `pcb-check io` to a `defn` with two sub-checks :
  - check for no connects
  - check for digital-io levels and constraints
- Extended digital-io check to handle parallel pull-up resistors and
  resistors not connected to rail voltages
- Extended digital-io check to check for pull-up resistors connected
  to different rail voltages
- Added helper functions for checking for DNP, no-connect and digital io pins
  to ocdb/checks

TODO: unit tests covering the known cases for the new checks

### DIFF
--- a/utils/checks.stanza
+++ b/utils/checks.stanza
@@ -11,6 +11,7 @@ defpackage ocdb/checks:
   import ocdb/property-structs
   import ocdb/design-vars
   import ocdb/tolerance
+  import lang-utils
 
 ; #CHECK(
 ;   condition = p-props is GenericPin,                    ; A boolean check                     (True|False)                       
@@ -27,26 +28,45 @@ defpackage ocdb/checks:
 ; ==== Check convenience functions ===
 ; ====================================
 
+
+doc:"Set of standard checks to run on a top-level-module. \
+                                                          \
+     - Component instances marked DNP are skipped         \
+     - Resistors, Capacitors are checked                  \
+     - Components with temperature rating are checked     \
+       against design variables, see ocdb/design-vars     \
+     - Pins are checked                                   \
+       - GenericPin properties are checked                \
+       - PowerPin properties are checked                  \
+       - ResetPin properties are checked                  \
+     - Landpattern checks are run, see check-landpattern  \
+     - Digital I/O pin checks are run, see check-io       "
 public defn check-design (module:JITXObject):
   inside pcb-module:
-    for i in component-instances(module) do :
-      if not (has-property?(i.DNP) or has-property?(i.do-not-populate) or has-property?(i.dnp)) :
-        check rated-temperature(i)
-        if has-property?(i.capacitance) :
-          check capacitor-component(i)
-        if has-property?(i.resistance) :
-          check resistor-component(i)
+    for i in populated-components(module) do :
+      check rated-temperature(i)
+      check capacitor-component(i) when has-property?(i.capacitance)
+      check resistor-component(i)  when has-property?(i.resistance)
+      
+      for p in pins(i) do : 
+        check generic-pin(i) when has-property?(p.generic-pin)
+        check power-pin(i)   when has-property?(p.power-pin)
+        check reset-pin(i)   when has-property?(p.reset-pin)
 
-        for p in pins(i) do :
-          if has-property?(p.generic-pin) :
-            check generic-pin(p)
-          if has-property?(p.power-pin) :
-            check power-pin(p)
-          if has-property?(p.reset-pin) :
-            check reset-pin(p)
-
-    check io(module)
+    check-io(module)
     check-all-landpatterns(module)
+
+; Checks if an object is DNP (do-not-populate)
+defn dnp? (obj:JITXObject) : 
+  has-property?(obj.DNP) or 
+  has-property?(obj.do-not-populate) or 
+  has-property?(obj.dnp)
+
+; Returns component instances not marked dnp?
+defn populated-components (module:JITXObject) -> Tuple<JITXObject> : 
+  component-instances(module) 
+    $> filter{{not dnp?(_)}, _}
+    $> to-tuple
 
 defn get-other-pin (comp:JITXObject,  p:JITXObject) :
   for ps in pins(comp) find! :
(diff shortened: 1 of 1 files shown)


utils/checks.stanza
```diff
@@ -113,8 +133,207 @@ public pcb-check rated-temperature (component:JITXObject):
...
+  val io-pins  = filter(is-digital-io-pin?, all-pins)
+  for p in io-pins do : 
+    val all-connected-pins = all-pins 
+      $> filter{{connected?([p, _0]) and _0 != p}, _}
+      $> to-tuple   
+    val connected-io-pins = all-connected-pins
+      $> filter{is-digital-io-pin?, _}
+      $> to-tuple
+    
+    #CHECK(condition =            not empty?(connected-io-pins),
```
bhusang, line 190: This check would seem to check if digital-output and digital-io pins are positively connected to at least one other digital-io pin. That check is not quite what we want as we should check and error only when more than 1 digital-output (not digital-io) pin is connected together. You can have a digital-output and digital-io pin 
1) unconnected
2) connected to each other (digital-output -> 1+ digital-io pins)
3) connected to 1+ digital-input pis

utils/checks.stanza
```diff
@@ -113,8 +133,207 @@ public pcb-check rated-temperature (component:JITXObject):
...
+    locators =             [io-pin])
+
+  ; Next, we check that all the pull-up resistors are connected to the same rail
+  val same-rail-voltage? = all-equal? $ 
+    for resistor in pull-up-resistors seq :
+      for p in pins(resistor) first! : 
+        property?(p.rail-voltage)
+    
+  #CHECK(
+    condition =            same-rail-voltage?,
```
bhusang, line 256: Can we add or replace this check with one that checks that all resistors are connected to the same net (vdd-like)? That is probably more germane to the check and allows for switchable rail nets. In other words, the (vdd) net for all resistors should be the same (similar to the rail-voltage property).

utils/checks.stanza
```diff
@@ -113,8 +133,207 @@ public pcb-check rated-temperature (component:JITXObject):
...
+
+  ; Here we find the value of the rail voltage
+  val rail-voltage = let : 
+    val r = pull-up-resistors[0]
+    for p in pins(r) first! :   
+      property?(p.rail-voltage)
+
+  ; And the total resistance of parallel connected
+  ; pull-up resistors
+  val total-resistance = 1.0 / sum $
```
bhusang, line 273: This is SO cool to calculate the parallel value of resistance using stanza syntax!

utils/checks.stanza
```diff
@@ -113,8 +133,207 @@ public pcb-check rated-temperature (component:JITXObject):
...
+    $> to-tuple
+  
+; Checks if a pin is marked "no-connect"
+defn no-connect? (p:JITXObject) -> True|False : 
+  has-property?(p.no-connect) or
+  has-property?(p.nc) or 
+  has-property?(p.NC)
+
+; Returns if a pin is a digital-io pin
+defn is-digital-io-pin? (p:JITXObject) -> True|False :
```
bhusang, line 150: We probably need to split this into a `is-digital-output-pin?` and `is-digital-io-pin?` as that would facilitate the checks for avoiding multiple outputs on the same net but allowing multiple io pins.

utils/checks.stanza
```diff
@@ -113,8 +133,207 @@ public pcb-check rated-temperature (component:JITXObject):
...
+; Returns all the pins that are electrically-meaningful, ie are not 
+; from components marked do-not-place
+defn all-pins (module:JITXObject) -> Tuple<JITXObject> : 
+  populated-components(module) 
+    $> seq-cat{pins, _}
+    $> to-tuple
+  
+; Checks if a pin is marked "no-connect"
+defn no-connect? (p:JITXObject) -> True|False : 
+  has-property?(p.no-connect) or
```
bhusang, line 145: So cool!

utils/checks.stanza
```diff
@@ -27,27 +28,46 @@ defpackage ocdb/checks:
...
-          if has-property?(p.generic-pin) :
-            check generic-pin(p)
-          if has-property?(p.power-pin) :
-            check power-pin(p)
-          if has-property?(p.reset-pin) :
-            check reset-pin(p)
-
-    check io(module)
+    for i in populated-components(module) do :
+      check rated-temperature(i)
```
bhusang, line 47: This is SO much better organized! Major kudos.

utils/checks.stanza
```diff
@@ -27,27 +28,46 @@ defpackage ocdb/checks:
...
+      check capacitor-component(i) when has-property?(i.capacitance)
+      check resistor-component(i)  when has-property?(i.resistance)
+      
+      for p in pins(i) do : 
+        check generic-pin(i) when has-property?(p.generic-pin)
```
bhusang, line 52: Ditto!

utils/checks.stanza
```diff
@@ -113,8 +133,207 @@ public pcb-check rated-temperature (component:JITXObject):
...
+    val connected-io-pins = all-connected-pins
+      $> filter{is-digital-io-pin?, _}
+      $> to-tuple
+    
+    #CHECK(condition =            not empty?(connected-io-pins),
+           name =                 "IO pin checks"
+           description =          "Check io pin properties"
+           category =             "pin-check"
+           subcheck-description = "Check that no-connect pads are not connected",
+           pass-message =         "Pin %_ is marked no-connect and is correctly not connected" % [ref(p)],
```
bhusang, line 195: The description needs to be tweaked :)

bhusang (changes requested):
@m-hilgendorf Looks pretty awesome. There were some issues around digital-output and digital-io that we need to update but I can also do that if you want.

utils/checks.stanza
m-hilgendorf: Can you clarify the 2nd case, is that multiple digital output pins connected to each other? I'm not sure if this check would ever fail if the pin can be unconnected or connected to any number of digital output/input/io pins. Would we fail if it was connected to any pins that are *not* digital i/o instead? 

This check was translated directly from the current implementation of `io`, the logic doesn't seem correct to me in either case (input pins are never checked at this stage).

utils/checks.stanza
m-hilgendorf: Sure, I'm looking into this. I'm not sure if we support net-level introspection for connected pins yet, but this is a good use case.

utils/checks.stanza
bhusang: More than one digital-output pin connected together is a no-no.
One digital-output can connect to 1 or more digital-io pins (because they can be used as input). 
One digital-output can be unconnected (and that is not an error)
Digital-output cannot be connected to power-supply-pins.

We do have to enhance the checks a bit which I want to do with your assistance (like checking digital-input pins)

utils/checks.stanza
bhusang: If we can't check the net itself, then can we check the pins on that net (vdd) for a PowerPin property?

utils/checks.stanza
m-hilgendorf: Fixed!

utils/checks.stanza
m-hilgendorf: Updated!

utils/checks.stanza
m-hilgendorf: Added!

utils/checks.stanza
m-hilgendorf: Ok, updated. The check is slightly more specific now : 

- All i/o pins are checked (using the function you described below)
- If the i/o pin is an output, check that it isn't connected to any other outputs

Let me know if this meets what we're after, I think the semantics of the original check are being changed slightly

bhusang (approved):
I am going to accept this PR as it is now but I will tweak the logic a little bit wrt digital-output and digital-io definitions.

m-hilgendorf:
Perfect, thanks!

PR #219, opened 2022-05-03
Added checks for power-states

m-hilgendorf:


bhusang (approved):
Looks good to me, some gnarly logic in power-state propagation which looks cool.

utils/property-structs.stanza
```diff
@@ -128,3 +128,77 @@ public pcb-struct ocdb/property-structs/CrystalResonator :
...
+      val instance = containing-instance(pin) as Instance
+      if has-property?(instance.power-supply-component) and 
+         property(instance.power-supply-component) : 
+        do(visit{_, power-states}, pins(instance))
+
+  ; propagate the power-states to connected power-pins
+  do(visit{_0, property(_0.power-states)}, roots)
+
+  ; propagate from power pins to io-pins
+  for io-pin in filter(is-digital-io?, all-pins) do :
```
bhusang, line 191: You need to add ocdb/checks to the package import list? Per CI, I think.

m-hilgendorf: Fixed! I moved the function definition to `ocdb/checks`, which makes more sense to me and doesn't expose a private API

bhusang (approved):
Looks good!

PR #223, opened 2022-05-05
Update mcu design to search for valid pin assignments

d-haldane:


utils/micro-controllers.stanza
```diff
@@ -71,6 +71,21 @@ public defn micro-controller (params:Tuple<KeyValue>) -> (Tuple<KeyValue<Symbol,
   val component = to-jitx(mcu)
   mcu-module(component)
 
+public defn find-micro-controller (core-parameters:Tuple<KeyValue<String, ?>>, 
+                            module-parameters:Tuple<KeyValue<Symbol, ?>>,  
+                            mcu-design) :
+  val possible-mcus = MicroControllers(core-parameters, 10)
```
bhusang, line 77: Why 10? Is that just a way to limit it to the 10 best? What if none of the 10 works in the pin search?

d-haldane: Then it fails. This design tries the 10 smallest given the parameters. Checking each one is quite slow. I could make the number to check an agrument if you want.

designs/mcu.stanza
```diff
@@ -10,13 +10,28 @@ defpackage ocdb/designs/mcu :
   import ocdb/design-vars
   import ocdb/micro-controllers
 
-println("Available mcu cores: %_" % [look-up-micro-controllers("core")])
-
 val BOARD-SHAPE = RoundedRectangle(25.0, 25.0, 0.25)
-pcb-module my-design:
-  inst mcu : micro-controller(["core" => "ARM Cortex-M3"])([`cap-bypass-package => 4.7e-6])
+DESIGN-QUANTITY = 0
+OPTIMIZE-FOR = ["area"]
```
PhilippeFerreiraDeSousa, line 15: FYI, you are still filtering on sellers in your jitx.params / user.params.

bhusang (approved):
Looks good!

utils/micro-controllers.stanza
m-hilgendorf: adding as a parameter defaulted to 10 is a good strategy

PhilippeFerreiraDeSousa: The default limit is 25 across the board
```
public defn dbquery-all (params: Tuple<KeyValue<String, Tuple<String>|Tuple<Double>|String|Double|Int>>) -> Tuple<JSON> :
  dbquery(params, 25)
```
However I haven't defined constructors without limit argument fro Resistors, Inductors, Capacitors, MicroControllers. But better not start mixing different values for default limits

PR #227, opened 2022-05-10
Fix for invalid check for pullup resistors on Digital I/O lines

m-hilgendorf:


bhusang (approved):
Looks good -- or versus and

PR #228, opened 2022-05-13
Connected pins introspection

PhilippeFerreiraDeSousa:
Checks performance improvements


utils/checks.stanza
```diff
@@ -179,34 +194,33 @@ pcb-check no-connect-pins-not-connected (module:JITXObject) :
...
-    val connected-out-pins = all-connected-pins 
-      $> filter{has-property?{_.digital-output}, _}
-      $> to-tuple
+pcb-check digital-io-levels (module:JITXObject) :
+  val io-pins = filter(is-digital-io?, all-populated-pins(module))
+
+  ; FIXME! Could we split them first in connected groups and then run the loop only once per connected group?
+  ; Issue:: some of the logic is assymetrical and considers this p as special in the connected group considered
+  for p in io-pins do :
+    val all-connected-pins = populated-connected-pins(p)
```
PhilippeFerreiraDeSousa, line 205: We can get all instances of a module and then build a cache of pin => dnp info. Then `populated-connected-pins` can filter on non-denp?

PhilippeFerreiraDeSousa: We need to create a hashset of visited pin and vector of connected groups

utils/checks.stanza
```diff
@@ -179,34 +194,33 @@ pcb-check no-connect-pins-not-connected (module:JITXObject) :
...
-            locators =             [p])
+             name =                 "IO pin checks"
+             description =          "Check io pin properties"
+             category =             "pin-check"
+             subcheck-description = "Check output pins are correctly connected",
+             pass-message =         "Pin %_ is marked as an output and not connected to any output pins" % [ref(p)],
+             fail-message =         "Pin %_ is marked as an output and connected to the following output pins: %," % [ref(p), seq(ref, connected-out-pins)],
+             locators =             [p])
+
+    val connected-io-pins = filter(is-digital-io?, all-connected-pins)
```
PhilippeFerreiraDeSousa, line 220: Rename `is-digital-io?` as `is-digital-pin?`.
For each connected group, only once we want to filter on io vs input vs output. If needed in some logic we can uses that input and output are subset of io

utils/checks.stanza
```diff
@@ -236,22 +252,26 @@ defn cmos-and-ttl-checks (out-p:JITXObject,
...
+    connected-to-rail?(instance)
+
+  ; FIXME!: A lot of pins must have the same containing instance!,
+  ; no need to recompute pull-up? and have a lot of checks done on the same instance
   ; First, we search for the pull-up-resistors that might be connected to this pin
-  val pull-up-resistors = to-tuple $
-    for p in all-connected-pins seq? :
-      val instance = containing-instance(p) as Instance
-      val connected-to-rail? = 
-        for p in pins(instance) any? :
```
PhilippeFerreiraDeSousa, line 246: Shadows `p` from parent scope

utils/checks.stanza
```diff
@@ -236,22 +252,26 @@ defn cmos-and-ttl-checks (out-p:JITXObject,
...
-                            all-pins:Tuple<JITXObject>) : 
+                            all-connected-pins:Tuple<JITXObject>) :
+  defn connected-to-rail? (instance: Instance) :
+    for p in pins(instance) any? :
+      ; Assumes that power states have been propagated.
+      has-property?(p.rail-voltage) or
+      has-property?(p.power-pin)
+
+  ; lazy 'and', connected-to-rail? can be heavy
+  defn pull-up? (instance) :
```
PhilippeFerreiraDeSousa, line 263: Cache the output

utils/checks.stanza
```diff
@@ -236,22 +252,26 @@ defn cmos-and-ttl-checks (out-p:JITXObject,
 ;
 defn open-collector-checks (io-pin:JITXObject,
```
PhilippeFerreiraDeSousa, line 253: Do not call this 100 times for a connected group with 100 io pins + we are missing a pin in the input (original p)

utils/checks.stanza
```diff
@@ -157,178 +223,269 @@ defn is-digital-input? (p:JITXObject) -> True|False :
...
+
+defn categorize-digital-pins (pins: Seqable<Pin>) -> DigitalPinCategories :
+  val ios = Vector<DigitalPin>()
+  val inputs = Vector<DigitalPin>()
+  val outputs = Vector<DigitalPin>()
+
+  for p in pins do :
+    ; TODO: Checking that rail-voltage propagation is correct for collector check, can remove later
+    if has-property?(p.power-pin) and not has-property?(p.rail-voltage) :
+      fatal("Power pin %_ does not have a rail voltage property" % [ref(p)])
```
bhusang, line 295: This should be converted to a `#CHECK()` and not a fatal.

PhilippeFerreiraDeSousa: This ensures correction of our rail voltage propagation algo, this is not checking anything from the user design.

PhilippeFerreiraDeSousa: If you think that this function may be called by a user forgetting to run the voltage propagation pass, we should throw an exception asking the user to run the voltage propagation pass, or better we should automatically detect the dependency and run the voltage propagation pass

designs/grid-resistors.stanza
```diff
@@ -0,0 +1,51 @@
+#use-added-syntax(jitx)
+defpackage ocdb/designs/grid-resistors :
+  import core
+
+  import jitx
+  import jitx/commands
+  import ocdb/defaults
+  import ocdb/generic-components
+
+val board-shape = RoundedRectangle(10.0, 10.0, 0.5)
```
bpellenard, line 10: 0.5 or 5.0?

utils/checks.stanza
bhusang: The issue is that the voltage propagation pass requires data that needs to be present in the design. We can't autorun the rail-voltage because we don't have apriori knowledge of the rail-voltage or power-supply-pin usage. I think that it's best to cast this as a #CHECK for now.

PR #229, opened 2022-05-18
extra checks in generic-components that are not needed (diode now included)

bhusang:
Removes extra checks that are not part of the check-design() flow. If you run check-design() then the resistor/capacitor/inductor pin connectivity checks are run by default.

### DIFF
--- a/utils/checks.stanza
+++ b/utils/checks.stanza
@@ -50,9 +50,10 @@ public defn check-design (module:JITXObject):
   inside pcb-module:
     for i in populated-components(module) do :
       check rated-temperature(i)
-      check capacitor-component(i) when has-property?(i.capacitance)
-      check resistor-component(i)  when has-property?(i.resistance)
-      check inductor-component(i)  when has-property?(i.inductance)
+      check capacitor-component(i) when has-property?(i.capacitor)
+      check resistor-component(i)  when has-property?(i.resistor)
+      check inductor-component(i)  when has-property?(i.inductor)
+      check diode-component(i)     when has-property?(i.diode)
       
       for p in pins(i) do :
         check generic-pin-check(p) when has-property?(p.generic-pin)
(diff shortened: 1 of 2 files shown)


electricfungi:
the new check connected check stops checking all other pins on a device once it finds an unconnected pin. Desired behavior: check each pin and report error for each unconnected pin on these components. There is currently no check connected for 2 pin LEDs.

electricfungi (changes requested):
the new check connected check stops checking all other pins on a device once it finds an unconnected pin. Desired behavior: check each pin and report error for each unconnected pin on these components. There is currently no check connected for 2 pin LEDs.

bhusang:
@electricfungi So I added a diode pin connectivity check. The behavior to stop the checks is deliberate and a result of how the checks system was setup. Will be changed when the check infrastructure is re-engineered but that is not the scope of this PR. You can attach a `NC` property to a pin that is intended to not be connected.

utils/checks.stanza
```diff
@@ -938,6 +939,27 @@ public pcb-check voltage-levels (p:JITXObject, range:[Double,Double,Double]) :
 ; ====================================
 ; public pcb-check pull-up-check (p:JITXObject) :
 
+public pcb-check diode-component (d:JITXObject) :
+  val name = "Diode checks"
+  val description = "Check diode component properties"
+  val category = "component-check"
+
+  ; should check that pins of the resistor are connected (if they don't have a NC property)
```
electricfungi, line 947: diode

utils/checks.stanza
```diff
@@ -959,13 +981,13 @@ public pcb-check inductor-component (l:JITXObject) :
       )
 
   #CHECK(
-    condition =            has-property?(l.inductor),
+    condition =            has-property?(l.inductance),
```
electricfungi, line 984: this will cause anything with an inductance property assigned to call this check vs an inductor. desired behavior is for only inductor components to be checked

utils/checks.stanza
```diff
@@ -1037,16 +1059,14 @@ public pcb-check resistor-component (r:JITXObject) :
         locators =             [r]
       )
 
-
-
   #CHECK(
-    condition =            has-property?(r.resistor),
+    condition =            has-property?(r.resistance),
```
electricfungi, line 1063: this will cause anything with a resistance property assigned to call this check vs an resistor. desired behavior is for only resistor components to be checked

utils/checks.stanza
```diff
@@ -1116,13 +1136,13 @@ public pcb-check capacitor-component (c:JITXObject) :
       )
 
   #CHECK(
-    condition =            has-property?(c.capacitor),
+    condition =            has-property?(c.capacitance),
```
electricfungi, line 1139: this will cause anything with an capacitance property assigned to call this check vs a capacitor. desired behavior is for only capacitor components to be checked.

electricfungi (changes requested):
Checked nc property skips check and that property can be set at component or module level. Checked inductors, resistors, caps, diodes for connectivity. Assigning properties like inductance to pins on things other than inductors (example-connectors) will be useful for high speed signals and checking the inductor property vs an inductance property will avoid unintentional checks here. For R/L/C/D modules with non-connected mounts, the nc property will need to be added to the component, but that's a rare case handled once at the component level with the nc property.

utils/checks.stanza
bhusang: Good catch

utils/checks.stanza
bhusang: Not quite. The check of the inducto `inductor-component()` is only triggered when the property `inductor` is defined (line 59):
```
      check inductor-component(i)  when has-property?(i.inductor)
```

utils/checks.stanza
bhusang: Same reason as above -- the actual code does not run unless the `resistor` property exists on the component in question.

utils/checks.stanza
bhusang: Same reason as above -- the actual code does not run unless the `capacitor` property exists on the component in question.

PR #231, opened 2022-05-20
Bg/skip operating point

bhusang:
@m-hilgendorf So the power states code almost works. The code that actually does the final equality check does not seem to be working:
```
public pcb-check power-states (module:JITXObject) :

  for connected-group in populated-connected-groups(module) do :
    val pin-categories = categorize-digital-pins(connected-group)
    val digital-pins = to-tuple $ cat-all([inputs(pin-categories),
                                           outputs(pin-categories),
                                           ios(pin-categories)])

    for digital-pin in digital-pins do :
      check-has-power-states $ pin(digital-pin)

    val condition? = all-equal? $
      for digital-pin in digital-pins seq :
        property(pin(digital-pin).power-states)

    #CHECK(condition    = condition?
            name        = "Power States"
            description = "Checks that all components have matching power states."
            category    = "power-states"
            subcheck-description = "Checking power states match on connected pins"
            pass-message = "Connected pins %, have matching power states" % [pin-refs(digital-pins)]
            fail-message = "Connected pins %_ and %_ do not have matching power states"  % [pin-refs(digital-pins)]
            locators = to-jitx-pins(digital-pins))
```
The all-equal condition is always true it appears even when the power-states are verifiably different.

### DIFF
--- a/utils/checks.stanza
+++ b/utils/checks.stanza
@@ -63,7 +63,6 @@ public defn erc-check-design (module:JITXObject):
         check generic-pin-check(p) when has-property?(p.generic-pin)
         check power-pin-check(p)   when has-property?(p.power-pin)
         check reset-pin-check(p)   when has-property?(p.reset-pin)
-
     check-io(module)
 
 public defn drc-check-design (module:JITXObject):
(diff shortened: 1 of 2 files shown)


utils/checks.stanza
```diff
@@ -47,20 +47,36 @@ Set of standard checks to run on a top-level-module.
...
   inside pcb-module:
     for i in populated-components(module) do :
       check rated-temperature(i)
-      check capacitor-component(i) when has-property?(i.capacitance)
-      check resistor-component(i)  when has-property?(i.resistance)
-      check inductor-component(i)  when has-property?(i.inductance)
+      check capacitor-component(i) when has-property?(i.capacitor)
+      check resistor-component(i)  when has-property?(i.resistor)
+      check inductor-component(i)  when has-property?(i.inductor)
+      check diode-component(i)     when has-property?(i.diode)
```
m-hilgendorf, line 60: Note: we should make sure that importers that attach these properties follow the new convention so checks are run on imported designs

utils/checks.stanza
```diff
@@ -1728,23 +1768,25 @@ public defn propagate-power-states (module:JITXObject) :
   ; DFS walk of the netlist, to update power-states of 
   ; connected power-pins 
   val visited = HashSet<JITXObject>()
-  defn visit (pin:JITXObject, power-states:Tuple<PowerState>) :
+  defn visit-pins (pin:JITXObject, power-states-tuple:Tuple<PowerState>) :
```
m-hilgendorf, line 1771: Style note: since `power-states` has the `Tuple` as an annotation already it doesn't need it to be annotated in the variable name

utils/checks.stanza
```diff
@@ -1728,23 +1768,25 @@ public defn propagate-power-states (module:JITXObject) :
...
+  defn visit-power-supply-components (pin:JITXObject, power-states-tuple:Tuple<PowerState>) :
+    val instance = containing-instance!(pin)
+    if has-property?(instance.power-supply-component) and
+        property(instance.power-supply-component) : 
+      do(visit-pins{_, power-states-tuple}, pins(instance))
 
   ; propagate the power-states to connected power-pins
-  do(visit{_0, property(_0.power-states)}, roots)
+  do(visit-pins{_0, property(_0.power-states)}, roots)
+  do(visit-power-supply-components{_0, property(_0.power-states)}, roots)
```
m-hilgendorf, line 1789: This change looks out of scope of the rest of the PR which only touches passives and OP checks, it may be contributing to the failure seen

m-hilgendorf (approved):
LGTM, one style note and a comment on the PS check changes which might be causing issues.

utils/checks.stanza
bhusang: Yes, I overloaded the PR in the rush to get it done. This was an attempt to fix another propagation issue.

utils/checks.stanza
bhusang: I looked in the generic component code and we do annotate the properties, but we should also check the importer code assuming we know the type of component that we are importing.

PR #233, opened 2022-05-24
Split off property propagation into new package

m-hilgendorf:
- Added tests for base cases in property propagation
- Added API for handling conflicts, custom property updates,
  and propagation across components/instances
- New impl of propagate-rail-voltage using new internals


utils/property-propagation.stanza
```diff
@@ -0,0 +1,220 @@
...
+  ;
+  ; Note: since we look at flattened nets, a pin may only be connected to one net.
+  val nets-by-pin = HashTable<JITXObject, JITXObject>()
+  
+  ; A table of nets that have the property we are looking for.
+  val relevent-nets = HashSet<JITXObject>()
+  
+  ; First, we extract the nets-by-pin and relevant-nets tables.
+  for net in unique-nets(root-module) do :
+    ; println("Indexing net (%_): %," % [net, seq(ref, pins-on-net(net))])
```
m-hilgendorf, line 69: Remove printers, out dated comments

utils/property-propagation.stanza
```diff
@@ -0,0 +1,220 @@
...
+    defn updater (p:JITXObject, name:Symbol, value:jitx/client/JITXValue) : 
+      set-property(p, `rail-voltage, value)
+    
+;==============================================================================
+;============================= Helpers ========================================
+;==============================================================================
+; TODO: refactor not to use undocumented internal APIs
+
+doc:"Returns if an object has a property."
+defn has-property? (obj:JITXObject, prop:Symbol) -> True|False :
```
m-hilgendorf, line 200: Add this function to internal API instead of hacking it through undocumented APIs

utils/property-propagation.stanza
```diff
@@ -0,0 +1,220 @@
...
+;==============================================================================
+; TODO: refactor not to use undocumented internal APIs
+
+doc:"Returns if an object has a property."
+defn has-property? (obj:JITXObject, prop:Symbol) -> True|False : 
+  ; TODO: introspection instead of using internal APIs explicitly
+  jitx/client/get-property?(jitx/commands/JITX-CLIENT, false, obj, prop) is One
+
+doc:"Returns if a pin is connected to a component marked DNP"
+defn dnp-pin? (pin:JITXObject) -> True|False :
```
m-hilgendorf, line 205: Use `ocdb/checks dnp?` to check if containing instance is DNP

bhusang: This would be good to do because we need to always filter DNP components, but it makes DNP a critical property that needs to be valid in the design (if used, that is). I think that the whole notion of DNP needs to be promoted to a more basic construct than a ephemeral property.

utils/property-propagation.stanza
bhusang: This is undocumented? Wow. As it is used liberally in checks, for example.

utils/property-propagation.stanza
bhusang: Cool.

utils/property-propagation.stanza
m-hilgendorf: Agreed. There's actually a good bit of nuance to this feature that I think it needs to be moved eventually into a core pass in the JITX compiler.

utils/property-propagation.stanza
m-hilgendorf: has-property?(thing.property) is what we use extensively, but has-property?(thing,`property) as defined here uses internals that aren't exposed normally

utils/property-propagation.stanza
bhusang: Yes, that would be good. Is it too much to do now?

m-hilgendorf: Yes, it requires a more flushed out design. It would be helpful to have this pass at the user level for now to identify the intended behaviors before building out something internally.

bhusang: Ok, makes sense.

utils/property-propagation.stanza
bhusang: Ok

tests/test-property-propagation.stanza
```diff
@@ -0,0 +1,112 @@
...
+
+  pcb-module main-module : 
+    inst m1 : Module
+    inst m2 : Module
+
+    net GND ()
+    net (m2.c[1].p[2], m1.c[1].p[2])
+    for n in 0 to num-instances do : 
+      net (m1.c[n].p[1], m2.c[n].p[1])
+      net (m1.c[n].p[2], m2.c[n].p[2])
```
bhusang, line 27: Will this just duplicate line 24 when n == 1?

tests/test-property-propagation.stanza
```diff
@@ -0,0 +1,112 @@
...
+      net (m1.c[n].p[2], GND)
+
```
bhusang, line 29: This also seems in conflict with line 24 and line 27? Or is your intent to connect all p[2] pins to GND?

utils/property-propagation.stanza
bhusang: Ahh, there are still commented out println()?

utils/property-propagation.stanza
```diff
@@ -0,0 +1,220 @@
...
+    val pins = to-tuple(filter({not dnp-pin?(_)}, pins-on-net(net)))
+    val [pins-with-property, pins-without-property] = 
+      fork-on-seq(pins, 
+        fn (ps) : 
+          to-tuple(filter(has-property?{_, property-name}, ps))
+        fn (ps) : 
+          filter({not has-property?(_, property-name)}, ps))
+
+    ; println("pins-with-prop: %," % [seq(ref, pins-with-property)])
+
```
bhusang, line 105: Leaving the commented out println() in?

tests/test-property-propagation.stanza
m-hilgendorf: Yea, but that's intentional (trying to check multiple net statements being flattened into one)

bhusang (changes requested):
Just some commented out println() to be cleaned up?

m-hilgendorf:
Yes, I'd also like to get some more tests in

tests/test-property-propagation.stanza
bhusang: Ok, cool.

utils/property-propagation.stanza
```diff
@@ -0,0 +1,220 @@
...
+    val pin-hash = connected-pins(net)
+      $> seq{hash, _}
+      $> lazy-qsort
+      $> to-tuple
+    val result = not visited[pin-hash]
+    ; within indented() : 
+    ;   println(result)
+    if result :
+      add(visited, pin-hash)
+    result
```
PhilippeFerreiraDeSousa, line 28: It seems that `connect-groups` from checks.stanza is doing what you want in a more optimal way (linear in the number of component pins and removes dnp pins before searching for connected nets). Between here and `pins-on-net` you are calling `connected-pins` potentially 100 times on the same net if the net is made up in esir with
```
pin a
for i in 0 to 100
  net (a res.p[i])
```
+ sorting is n log(n) making the algo possibly non-linear + dnp pins are included in this n

m-hilgendorf: I do not want to reinvent introspections, but go directly to the plugin.

PhilippeFerreiraDeSousa: Here is the logic there
```
defn connect-groups (module:JITXObject) -> Connections :
  val visited-connections = HashSet<Pin>()
  val connected-groups = Vector<Tuple<Pin>>()
  val non-connected-pins = Vector<Pin>()

  for p in all-populated-pins(module) do :
    ; p is not added to visited-connections because we already looped over it, so it is clear.
    if visited-connections[p] :
      None()
    else :
      val connections = connected-pins(p)
      if empty?(connections) :
        add(non-connected-pins, p)
      else :
        add-all(visited-connections, connections)
        val connected-group = to-tuple $ cat([p], connections)
        add(connected-groups, connected-group)

  Connections(to-tuple(connected-groups),
              to-tuple(non-connected-pins))

; Returns all the pins that are electrically-meaningful, ie are not
; from components marked do-not-place
defn all-populated-pins (module:JITXObject) -> Tuple<Pin> :
  populated-components(module)
    $> seq-cat{pins, _}
    $> to-tuple ; FIXME! There can be a lot, do not to-tuple?
    $> {_ as Tuple<Pin>}

; Returns component instances not marked DNP
defn populated-components (module:JITXObject) -> Tuple<JITXObject> : 
  component-instances(module) 
    $> filter{{not dnp-component?(_)}, _}
    $> to-tuple
```

m-hilgendorf: And like the TODO mentions, this needs to be added as an introspection since it's redoing the work of flattening nets.

PhilippeFerreiraDeSousa: Unsure what you mean? In here it seems you can make the loop body linear in the length of the net without sorting:
```
doc: "Returns a list of unique nets within a module (nets that have a  \
      unique set of pins associated with them)"
public defn unique-nets (module:JITXObject) -> Seq<JITXObject> : 
  val visited = HashSet<Pin>()
  for net in nets(module) filter : 
    val pins = connected-pins(net)
    val result = not empty?(pins) and not visited[pins[0]]
    if result :
      add-all(visited, pins)
    result
```

m-hilgendorf: If the order of the pins returned by `connected-pins` (btw: should be `pins-on-net`, that's a bug) is different the hash will be different.

m-hilgendorf: I wouldn't focus on this too much - it will be moved to the compiler to work on flattened nets (instead of using net statements)

PhilippeFerreiraDeSousa: Yes, adding an introspection to return the connected nets directly is easy and this code will be replaced. Actually if there were to be a lot of dnps, `connected-group` from check.stanza could actually be quicker than an introspection giving out the connected nets and then filtering out dnps

PhilippeFerreiraDeSousa: > If the order of the pins returned by connected-pins (btw: should be pins-on-net, that's a bug) is different the hash will be different.
There is no manually hashing needed

PR #235, opened 2022-06-09
Updates for new activation flow

d-haldane:


designs/design-template.stanza
```diff
@@ -0,0 +1,49 @@
+#use-added-syntax(jitx)
+defpackage design-template :
+  import core
+  import jitx
+  import jitx/commands
+
+; =========================
+; Module to run as a design
+; ==========================
+val board-shape = RoundedRectangle(30.0, 18.5, 0.25)
```
m-hilgendorf, line 10: Normally in idiomatic stanza we would notate global constants as `BOARD-SHAPE`, do we want to keep the same style here?

designs/design-template.stanza
```diff
@@ -0,0 +1,49 @@
...
+pcb-module my-design :
+
+  inst version-label  : ocdb/board-text/version-silkscreen("REV 0.0.0")
+  place(version-label) at loc(0.0, height(board-shape) / 2.0 - 1.0) on Bottom
+  
+; ========================================
+; Configure the design and view the result
+; ========================================
+defn run-design (circuit:Instantiable, outline:Shape) :
+  set-design("CAD")
```
m-hilgendorf, line 20: Some comments here might be useful : 

```
  set-design("CAD") ; name our design, which creates an output directory of the same name.

  set-main-module(circuit) ; declare the "main" module, analogous to a root schematic sheet (phrasing?)
  set-rules(ocdb/rules/jjcpcb-rules) ; etc 
```

designs/design-template.stanza
```diff
@@ -0,0 +1,49 @@
...
+; ===================================
+; Update CAD, keeping layout progress
+; ===================================
+defn update-design () :
+  set-export-board?(false)
+  export-design()
+
+; ============
+; Run the code
+; ============
```
m-hilgendorf, line 48: I think this is a very clean addition that answers, "what does it mean to `run` a design?"

manufacturers/stackups.stanza
```diff
@@ -20,6 +20,11 @@ pcb-material prepreg :
   dielectric-coefficient = 4.26
   name = "PR4 Prepreg 2113/2116"
 
+pcb-material prepreg-2313 :
+  type = Dielectric
+  dielectric-coefficient = 4.05
```
m-hilgendorf, line 25: Is it useful to annotate units here, or is it obvious what they are?

m-hilgendorf (commented):
Super clean and easy to approach "hello world" example, I think. Just a handful of questions.

designs/design-template.stanza
```diff
@@ -0,0 +1,49 @@
...
+  
+; ========================================
+; Configure the design and view the result
+; ========================================
+defn run-design (circuit:Instantiable, outline:Shape) :
+  set-design("CAD")
+
+  set-main-module(circuit)
+
+  set-board(ocdb/defaults/default-board(ocdb/stackups/jlcpcb-jlc2313, outline))
```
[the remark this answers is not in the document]
d-haldane: Yeah, I think so - I was trying to strike a balance between verbosity, and showing how rules/stackups are things you can change.

manufacturers/stackups.stanza
d-haldane: This one is pretty definitive. Also unitless in this instance :P

designs/tutorial.stanza
```diff
@@ -109,5 +113,5 @@ defn run-design (circuit:Instantiable) :
   view-board()
   view-schematic()
 
-; run-design(simple-lm317a-regulator(10.0))
-run-design(regulator-fixture)
+run-design(simple-lm317a-regulator(12.0))
+; run-design(regulator-fixture)
```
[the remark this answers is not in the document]
d-haldane: Nope! let me change that.

designs/design-template.stanza
m-hilgendorf: Sorry, global in the sense that it is defined at the top level of a package. It may be preferable to move it to the body of `run-design` if it's intended to be local

manufacturers/stackups.stanza
m-hilgendorf: TIL!

designs/design-template.stanza
d-haldane: Usability wise, it helps to have the shape next to the definition of the `pcb-module` that will go in that design. I tried it inside the body of `run-design` but did not like the resulting legibility.

designs/design-template.stanza
d-haldane: I'm going to change `set-design` to `set-design-directory` instead of adding a comment.

m-hilgendorf (approved):

PR #239, opened 2022-06-20
Feature/route usb light

electricfungi:
Route USB light so we can export and order

### DIFF
--- a/designs/usb-light.stanza
+++ b/designs/usb-light.stanza
@@ -16,6 +16,17 @@ defpackage ocdb/designs/usb-light :
 
 val BOARD-SHAPE = RoundedRectangle(25.0, 25.0, 0.25)
 
+defn get-pad-pose (obj:JITXObject, pad-ref:String) : 
+  get-pad-pose(obj, pad-ref, false) 
+
+defn get-pad-pose (obj:JITXObject, pad-ref:String, pad-num:Int|False) -> Pose: 
+  val landpattern-pads = pads( landpattern(instance-definition(obj)) )
+  match(pad-num) :
+    (pad-num:Int) :
+      pose(find!({ref(_) == Ref(pad-ref)[pad-num]}, landpattern-pads))
+    (pad-num:False) :
+      pose(find!({ref(_) == Ref(pad-ref)}, landpattern-pads))
+
 defn led-circuit-resistance (vbus:Double, forward-voltage:Double, current:Double, num-leds:Int) -> Double :
   (vbus - (forward-voltage * (to-double(num-leds) )) ) / current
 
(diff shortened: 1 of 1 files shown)


designs/usb-light.stanza
```diff
@@ -31,15 +43,75 @@ pcb-module main-board:
...
+    val net-id2 = make-net( to-symbol(net-name2), [(res[i].p[1])])
+    net (net-id2 led[i + 4].c)
+
     net (gnd res[i].p[2])
-    place(led[i + 4]) at loc(-2.5, 5.0 * to-double(i) - 7.5, 90.0) on Top
-    place(led[i]) at loc(-7.5, 5.0 * to-double(i) - 7.5, 90.0) on Top
-    place(res[i]) at loc(2.5, 5.0 * to-double(i) - 7.5, 90.0) on Top
-  place(conn) at loc(10.75, 0.0, 90.0) on Top
+    ; place p5v track stub
+    val led-5v-x1 = x(led-i-point) - y(center(get-pad-pose(((led[i])), "a")))
```
fgreen, line 68: This (and basically every point computed below) looks like it's subtracting an x coordinate from a y coordinate. This is usually an indicator that something is wrong. I'm assuming this design works, but I'm suspicious that this would break if you did something like change the spacing between rows. Do you know why you need to mix coordinates here?

designs/usb-light.stanza
```diff
@@ -31,15 +43,75 @@ pcb-module main-board:
...
+    println("%_ %_ %_ " % [led-5v-x1 x(led-i-point)  y(center(get-pad-pose(((led[i])), "a"))) ])
+    val led-5v-y1 = y(led-i-point) 
+    geom(p5v) :
+      copper(LayerIndex(0)) = Line(trace-thickness, [ Point(led-5v-x1, led-5v-y1), Point(led-5v-x1 - 2.0, led-5v-y1) ])
+    ; place p5v track 
+      if i != 0 :
```
fgreen, line 74: Tiny nitpick, but using `i == 0` here and swapping the two clauses here might make this a tiny bit clearer. As a general rule it's better to avoid negation.

designs/usb-light.stanza
```diff
@@ -31,15 +43,75 @@ pcb-module main-board:
...
+        copper(LayerIndex(0)) = Line(trace-thickness, [Point(led-5v-x1 - 2.0, led-5v-y1) Point(led-5v-x1 - 2.0, led-5v-y1 - 5.0) ])      
+      else :
+        copper(LayerIndex(0)) = Polyline(trace-thickness, [Arc(Point(led-5v-x1 - 2.0, led-5v-y1), Point(-9.5, -9.5), Point(-8.0, -10.0))])
+
+    ; place track between the LEDs
+    val led-5v-x2 = x(led-i-point) - y(center(get-pad-pose(((led[i])), "c")))
+    val led-middle-x1 = x(led-i4-point) - y(center(get-pad-pose(((led[i + 4])), "a")))
+    println("led-5v-x2 %_" % [led-5v-x2])
+    geom(net-id1) :
+      copper(LayerIndex(0)) = Line(trace-thickness, [ Point(led-5v-x2, led-5v-y1), Point(led-middle-x1, led-5v-y1) ])
```
fgreen, line 84: Probably could add this definition earlier in the file: 
```
defn pad-center(part:JITXObject, pad-name:String) -> Point:
    center(get-pad-pose(part, pad-name))
```
Then this ought to become something like:
```suggestion
      copper(LayerIndex(0)) = Line(trace-thickness, [ pad-center(led[i], "c"), pad-center(led[i+4], "a")] )
```
But the coordinate mixing issue makes me wonder if something is going wrong here.

designs/usb-light.stanza
```diff
@@ -31,15 +43,75 @@ pcb-module main-board:
...
+        copper(LayerIndex(0)) = Line(trace-thickness, [Point(5.0, led-5v-y1) Point(5.0, led-5v-y1 - 5.0) ])
+  val conn-point = Point(10.75, 0.0)
+  place(conn) at loc(conn-point, 90.0) on Top
+  val conn-pin-x = x(conn-point) - y(center(get-pad-pose(((conn)), "p",1)))
+  val conn-gnd-y = y(conn-point) + x(center(get-pad-pose(((conn)), "p",5)))
+  val conn-p5v-y = y(conn-point) + x(center(get-pad-pose(((conn)), "p",1)))
+  geom(p5v) :
+    copper(LayerIndex(0)) = Line(trace-thickness, [ Point(5.7, conn-p5v-y), Point(conn-pin-x, conn-p5v-y) ])
+    copper(LayerIndex(0)) = Line(trace-thickness, [ Point(5.7, -8.0), Point(5.7, conn-p5v-y) ])
+    ; copper(LayerIndex(0)) = Line(trace-thickness, [ Point(5.7, -10.0), Point(-8.0, -10.0) ])
```
fgreen, line 106: I'd remove this or explain why it's here
```suggestion
```

designs/usb-light.stanza
```diff
@@ -31,15 +43,75 @@ pcb-module main-board:
...
+          copper(LayerIndex(0)) = Polyline(trace-thickness, [Point(-8.0, -10.0), Arc(Point(4.2, -10.0), Point(5.4, -9.0), Point(5.7, -8.0))])
+
+  geom(gnd) :
+    copper(LayerIndex(0)) = Line(trace-thickness, [ Point(5.0, conn-gnd-y), Point(conn-pin-x, conn-gnd-y) ])
+    copper-pour(LayerIndex(0), isolate = 0.1, rank = 1) = RoundedRectangle(6.0, 6.5, 1.0, loc(conn-point,0.0))
+    ; place(copper-pour(LayerIndex(0), isolate = 0.1, rank = 1) = RoundedRectangle(5.0, 7.0, 1.0)) at loc(conn-point,0.0) on Top
```
fgreen, line 112: I'd remove this or explain why it's here
```suggestion
```

fgreen (commented):
I left some comments about small things that might improve readability.

My only real concern is that x and y coordinates are being subtracted from each other. That's usually not a good sign so I'm wondering what's going on there. I'd like to try to run this design myself and see if I can see what's going on.

designs/usb-light.stanza
electricfungi: I agree a higher order function would be desirable. But it belongs in utils. This proposed function doesn't address the coordinate mixing issue. I'd like to put off developing higher order functions for another PR, maybe after we've identified a few more cases.

designs/usb-light.stanza
electricfungi: removed

designs/usb-light.stanza
electricfungi: removed

designs/usb-light.stanza
```diff
@@ -31,15 +43,75 @@ pcb-module main-board:
   val vbus-voltage = (nom-value(recommended-voltage(property(conn.VCC.power-pin)))) as Double
   val exact-resistance = led-circuit-resistance(vbus-voltage, property(led[0].forward-voltage), property(led[0].test-current), 2)
   inst res : gen-res-cmp(closest-std-val(exact-resistance,5.0), "0603")[4]
+  val trace-thickness = 0.254
   for i in 0 to 4 do :
```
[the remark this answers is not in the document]
electricfungi: updated

designs/usb-light.stanza
```diff
@@ -31,15 +43,75 @@ pcb-module main-board:
...
+    ; place parts
+    val led-i4-point = Point(-2.5, 5.0 * to-double(i) - 7.5)
```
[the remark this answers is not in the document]
electricfungi: changed to left and right

designs/usb-light.stanza
electricfungi: Yea this looks a little funny. get-pad-pose returns the pad pose in relation to the center of the component. Here, the components are placed rotated 90 degrees. So to find the shift X in relation to the pad to the board, I take the shift y of the pad to the component then apply that to the x of the component to the board. I think these kinds of functions highlight an opportunity to build higher order functions for this type of behavior.

designs/usb-light.stanza
electricfungi: ok

fgreen (approved):

PR #240, opened 2022-06-22
Feature/add esp32 wroom32

electricfungi:
add esp32-wroom-32x

### DIFF
--- a/components/espressif/esp32-wroom-32.stanza
+++ b/components/espressif/esp32-wroom-32.stanza
@@ -1,0 +1,200 @@
+; When renaming the stanza package name below, also rename it in the folder stanza.proj
+; See docs.jitx.com for help
+#use-added-syntax(jitx)
+defpackage ocdb/espressif/esp32-wroom-32:
+  import core
+  import collections
+  import jitx
+
+  import jitx/commands
+  import ocdb/defaults
+  import ocdb/land-patterns
+  import ocdb/tolerance
+  import ocdb/box-symbol
+  import ocdb/bundles
+  import ocdb/land-protrusions
+  import ocdb/design-vars
+
+public pcb-enum ocdb/espressif/esp32-wroom-32/esp32-wroom-package-type :
+  esp32-type-E
+  esp32-type-UE
+
+pcb-pad esp32-edge :
+  smd-pad(0.85, 0.9)
+
+pcb-pad esp32-middle :
+  smd-pad(0.9, 0.9)
+
+pcb-landpattern esp32-wroom (package:ocdb/espressif/esp32-wroom-32/esp32-wroom-package-type) :
+  val num-pins = 38                 ; number of pins of the component
+  val pitch = 1.27                 ; pitch of the pins of the component
+  val lead-span = tol(18.0,0.15)  ; the overall length of the component, from terminal-edge to terminal-edge
+  val package-x = 18.0    ; the x of the package 
+  var package-y = 19.2     ; the y of the package
+  val terminal-x = tol(0.85,0.0)   ; the x of the terminals
+  val terminal-y = tol(0.9,0.0)    ; the y of the terminals
+  val lead-fillets = lead-fillets(LeadlessConcaveCastellated, DENSITY-LEVEL)
+  var offset = 0.0
+  if (package is esp32-type-E) : 
+    package-y = 25.5
+    offset = (25.5 - 19.2) / 2.0
+  inside pcb-landpattern:
+    ;----------------------------------------------------
+    ; Compute adjustments to dimensions using IPC formula
+    val ipc = ipc-formula(
+      lead-span, terminal-x, terminal-y, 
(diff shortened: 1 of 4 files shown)


components/espressif/esp32-wroom-32.stanza
```diff
@@ -0,0 +1,277 @@
...
+
+  var package-string = "UE"
+  var package-desc = ", UMCX"
+  if pcb-antenna? :
+    package-string = "E" 
+    package-desc = ", PCB Antenna"
+
+  var temp-text = "N"
+  if high-temp:
+    if memory == 4 :
```
m-hilgendorf, line 46: Normally for cases like this we'd prefer to have an `ensure-...` function that is called first at the top of the function to validate the parameters of the generator, since it can be difficult to validate the conditions in one place, which is really easy to check (and extend).

components/espressif/esp32-wroom-32.stanza
```diff
@@ -0,0 +1,277 @@
...
+  description = string-join(["Module, ESP32, "package-desc mem-text ", WiFi 802.11b/g/n, Bluetooth v4.2 +EDR, Class 1, 2 and 3 Transceiver, Surface Mount"])
+  manufacturer = "Espressif" 
+  reference-prefix = "U"
+
+  pin EN 
+  port IO : pin[[0 2 4 5 12 13 14 15 16 17 18 19 21 22 23 25 26 27 32 33 34 35]]
+  pin RXD0 
+  pin TXD0 
+
+  pin-properties :
```
m-hilgendorf, line 65: It's been awhile since I've done this - is it valid to mix pin definitions and pin properties like this? (or necessary?)

components/espressif/esp32-wroom-32.stanza
```diff
@@ -0,0 +1,277 @@
...
+      require io:io-pin[2] from esp32
+      ocdb-uart.tx => io[0].p
+      property(io[0].p.digital-output) = DigitalOutput(CMOSOutput(min-max(0.0, 0.1 * 3.3), min-max(0.8 * 3.3, 3.3)), false, esp32.P3V3, esp32.GND)
+      ocdb-uart.rx => io[1].p
+      property(io[1].p.digital-input) = DigitalInput(0.25 * 3.3, 0.75 * 3.3, esp32.P3V3, esp32.GND, 50.0e-9)  
+
+  schematic-group(self) = ESP32
+
+; ================================================== Land Pattern ==================================================
+pcb-landpattern esp32-wroom (pcb-antenna?:True|False) :
```
m-hilgendorf, line 200: Is there a doc link for the package? Just curious if this is a weird one that the generators don't help with

utils/land-patterns.stanza
```diff
@@ -1132,7 +1132,7 @@ public pcb-landpattern smd-testpoint-pkg (d:Double) :
 
 ; Helper struct to contain the results of the IPC formula in section 3 of 
 ; 7351-B
-defstruct IpcResults:
+public defstruct IpcResults:
```
m-hilgendorf, line 1135: :+1:

utils/land-protrusions.stanza
```diff
@@ -26,6 +26,7 @@ public pcb-enum ocdb/land-patterns/leads/LandProtrusionType:
   SmallRectangularLeads,  ; Leads for rectangular components (chip capacitors, resistors, inductors) smaller than 0601
   BigRectangularLeads,    ; Leads for rectangular components (chip capacitors, resistors, inductors) 0601 or larger
   CylindricalLeads,       ; Cylindrical end cap leads
+  LeadlessConcaveCastellated, ; Lead less chip carrier leads
```
m-hilgendorf, line 29: Nice!

m-hilgendorf (approved):
A handful of style comments and one question that I think I know the answer to, otherwise looks great!

components/espressif/esp32-wroom-32.stanza
electricfungi: https://www.espressif.com/sites/default/files/documentation/esp32-wroom-32e_esp32-wroom-32ue_datasheet_en.pdf
this is where I can use the most help

m-hilgendorf: Yea that's what I figured, it's a weird one that doesn't fit the existing generators well. I don't think we have a working custom LGA (which would be the closest to this example) that will work. I think reusing the internals is appropriate until we can design a more generic generator that can handle these cases, but it's not obvious if that's possible.

components/espressif/esp32-wroom-32.stanza
electricfungi: assigning IO explicitly is required, but I removed the others

components/espressif/esp32-wroom-32.stanza
electricfungi: that has more lines of code in this case.

PR #243, opened 2022-06-24
added rated-temperature for led-display project

bhusang:
Added rated-temperature() to some components to enable erc checks to pass on the led-display project


electricfungi (approved):

electricfungi (changes requested):
I think the land pattern only parts need a min max temp range that won't fail for some design values. -40 to 125 is too tight

bhusang:
The temperature limits for standard FR4 are -50 to +135. I increased the tag-connect footprints to those new limits. This issue is interesting because if you used a different substrate, the limits would be different so we need some way to capture the substrate Tg.

electricfungi (approved):

PR #245, opened 2022-07-01
Design vars for set bom vendors and set design quantity in commandsstanza

PhilippeFerreiraDeSousa:
Goes with:
- https://github.com/JITx-Inc/jitx-client/pull/1178
- https://github.com/JITx-Inc/design-template/pull/9

### DIFF
--- a/utils/design-vars.stanza
+++ b/utils/design-vars.stanza
@@ -17,7 +17,6 @@ public var OPERATING-TEMPERATURE = [0.0 25.0]
 public var OPTIMIZE-FOR: Tuple<String> = ["area"]
 public var MAX-Z = 500.0
 public var MIN-PKG = "0402"
-public var DESIGN-QUANTITY = 100
 public var PREFERRED-MOUNTING = "smd" ; values in ["smd", "through-hole"]
 public var MIN-CAP-VOLTAGE = 10.0 ; Minimum voltage to allow for capacitors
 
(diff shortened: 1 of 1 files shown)


utils/design-vars.stanza
```diff
@@ -30,5 +29,16 @@ public var MAX-HOLE-SIZE-TOLERANCE:  Double = 0.0508 ; The tolerance on the diam
...
 public var HOLE-POSITION-TOLERANCE:  Double = 0.0508 ; The tolerance on placing any hole
 
-; For consolidation
-;public var PREFERRED-MANUFACTURERS = ["Yageo", "Panasonic Electronic Components", "Vishay Dale", "TE Connectivity Passive Product"]
+;============================================
+;====== BOM Generator configuration =========
+;============================================
+
+; Set it to false to allow any distributor
+public var APPROVED-DISTRIBUTOR-LIST = ["Allied Electronics & Automation"
```
ptuckey, line 37: Duncan mentioned yesterday that this list should be ordered by preference (see [JITX-2040](https://linear.app/jitx/issue/JITX-2040/bom-sort-by-preferred-vendors)).

1. Alphabetical looks pleasing to the eye! But, is there a preferred ordering to these we should be using in the default?
2. The name change from "preferred" to "approved" suggests we're now using as a filter (rather than allowing fallbacks to distributors outside of the list). Is that intended? (if so, we can match that in [JITX-2040](https://linear.app/jitx/issue/JITX-2040/bom-sort-by-preferred-vendors))

ptuckey (approved):

PR #246, opened 2022-07-01
fix run-final-passes

PhilippeFerreiraDeSousa:
- `run-final-passes` was incorrect, `assign-pins()` was reusing the `intermediate-design` from the start instead of the result of the 3 iteration of `propagate-rail-voltage` and `run-evals`
- `calculate-operating-point` was suboptimal
- add gnd prop to ble-mote to run the logic in CI

Goes with https://github.com/JITx-Inc/jitx-client/pull/1180

Tested with prints:
Screenshot from 2022-07-01 13-11-39

Using a net as gnd property:
Screenshot from 2022-07-01 14-18-56

Screenshot from 2022-07-01 13-10-22

With component pin:
Screenshot from 2022-07-01 13-12-51
Screenshot from 2022-07-01 13-15-53

With module pin:
Screenshot from 2022-07-01 13-16-43
Screenshot from 2022-07-01 13-17-43


d-haldane (approved):

PhilippeFerreiraDeSousa:
Thank you!

PR #248, opened 2022-07-08
Change how voltages are defined in designs

d-haldane:


utils/generator-utils.stanza
```diff
@@ -109,65 +109,39 @@ public defn connect-gnd-pins ():
...
-              property(inst.operating-point) = capacitor-operating-point(voltage)
-    else :
-      gnd-type-error!(self, gnd) when gnd is-not False
+    for i in component-instances(self) do :
+      if has-property?(i.resistance) or has-property?(i.capacitance) or has-property?(i.inductance):
+        if length(pins(i)) == 2 :
+          if has-property?(pins(i)[0].net-voltage) and has-property?(pins(i)[1].net-voltage) :
+            val p1 = property(pins(i)[0].net-voltage)
+            val p2 = property(pins(i)[1].net-voltage)
+            val voltage = abs(max((max-value(p1) - min-value(p2)), (max-value(p2) - min-value(p1))))
```
PhilippeFerreiraDeSousa, line 140: It seems the maximum of the 2 will always be positive. Actually, should we have a constructor check that ensures `tol-` and `tol+` are always positive? It seems it is very necessary and when I saw `Toleranced(Double, Double, Double)` I thought it was the usual `(min, norm, max)` but it is not

designs/can-stm32.stanza
```diff
@@ -58,7 +58,8 @@ pcb-module first-design :
   symbol(P3V3) = ocdb/symbols/supply-sym
   symbol(GND) = ocdb/symbols/ground-sym
 
-  property(self.gnd) = GND
+  property(GND.net-voltage) = ocdb/tolerance/typ(0.0)
+  ; property(P3V3.net-voltage) = ocdb/tolerance/typ(3.3)
```
PhilippeFerreiraDeSousa, line 62: Are we leaving this in comment?

designs/ble-mote.stanza
```diff
@@ -153,6 +153,11 @@ pcb-module sensor-mote :
   net D- (usb.usb-2.data.N)
   net D+ (usb.usb-2.data.P)
 
+  property(VBUS.net-voltage) = min-max(4.8, 5.2)
+  property(P5V0.net-voltage) = min-max(4.9, 5.1)
+  property(P3V3.net-voltage) = typ(3.3)
+  property(gnd.net-voltage) = typ(0.0)
```
PhilippeFerreiraDeSousa, line 159: You can remove `property(self.gnd) = gnd` that I added line 57 of this file

utils/checks.stanza
```diff
@@ -440,25 +439,25 @@ defn check-a-pull-resistor-exists (open-collectors: Tuple<DigitalPin>, pull-up-r
...
+defn check-and-return-unique-net-voltage (net-voltages: Tuple<Toleranced>, pull-up-resistors: Tuple<Instance>, open-collectors: Tuple<DigitalPin>) -> Toleranced :
   #CHECK(
-    condition =            all-equal?(rail-voltages),
+    condition =            all-equal?(net-voltages),
     name =                 "IO pin checks"
     description =          "Check io pin properties"
-    category =             "pin-check"
+    category =             "Pin Checks"
     subcheck-description = "Check that io pins marked OpenCollector have pull-up resistors connected to the same rail voltage",
     pass-message =         open-collector-pass-message(open-collectors, "all pull-up resistors with the same rail voltage"),
```
PhilippeFerreiraDeSousa, line 455: Messages still have the term `rail-voltage`, but can we still have the same logic like this? pull-up resistors need to be attached to supply rails.

utils/checks.stanza
```diff
@@ -381,34 +380,34 @@ defn cmos-and-ttl-checks (cmos-ttl-outputs:Tuple<DigitalPin>,
 ;
 defn open-collector-checks (open-collectors:Tuple<DigitalPin>,
                             pin-categories:DigitalPinCategories) :
-  val [pull-up-resistors, rail-voltages] = collect-pull-up-resistors-and-their-rail-voltages(open-collectors, pin-categories)
+  val [pull-up-resistors, net-voltages] = collect-pull-up-resistors-and-their-net-voltages(open-collectors, pin-categories)
   check-a-pull-resistor-exists(open-collectors, pull-up-resistors)
```
PhilippeFerreiraDeSousa, line 384: The previous `propagate-rail-voltage` was finding supply powers and then propagating them on with rail-voltage properties. So replacing by `net-voltages` is incorrect. Every connected component pin is supposed to have a net voltage now so every resistor is detected as pull-up resistor and the check on pull-up resistor never fails

PhilippeFerreiraDeSousa (commented):
Hey, I left a few comments

modules/power-regulators.stanza
```diff
@@ -57,8 +59,10 @@ for v in tlv do:
       schematic-group(ps) = power
       ;Create nets
       net (ps.gnd, neg(input), neg(output))
-      net (ps.vin, pos(input), ps.en)
-      net (ps.vout, pos(output))
+      net vin (ps.vin, pos(input), ps.en)
+      net vreg (ps.vout, pos(output))
+      property(vreg.net-voltage) = ocdb/tolerance/typ(voltage(output))
+      property(vin.net-voltage) = ocdb/tolerance/typ(voltage(input))
```
bhusang, line 65: Why would you not assign 0.0v to the ground here?
net gnd (ps.gnd neg(input), neg(output))

property(gnd.net-voltage) = ocdb/tolerance/typ(0.0)

utils/checks.stanza
bhusang: If the user sets a voltage on the resistor pin that does not satisfy the vih of the input (after calculation), then the error will be triggered. We do calculate the current through the pull-up resistor and should use that current * resistance subtracted from the net-voltage to determine the input voltage.

utils/checks.stanza
```diff
@@ -483,22 +482,22 @@ defn compute-equivalent-resistance (pull-up-resistors: Tuple<Instance>) :
       val resistance = check-pull-up-resistor-does-not-have-0-ohm-resistance(resistor)
       1.0 / resistance
 
-defn check-sink-current-meets-collector-ratings (rail-voltage: Double, pull-up-resistors: Tuple<Instance>, open-collectors: Tuple<DigitalPin>) :
+defn check-sink-current-meets-collector-ratings (net-voltage: Toleranced, pull-up-resistors: Tuple<Instance>, open-collectors: Tuple<DigitalPin>) :
   val equivalent-resistance = compute-equivalent-resistance(pull-up-resistors)
   ; Next, compute the sink current (V / R) and check if it is in spec for the driver.
-  val sink-current = rail-voltage / equivalent-resistance
+  val sink-current = max-value(net-voltage) / equivalent-resistance
```
bhusang, line 488: This needs to be modified a bit?

val sink-current = (max-value(net-voltage) - min-value(vol(driver))) / equivalent-resistance

which will calculate the maximum sink-current and check it against the driver spec.

utils/generator-utils.stanza
```diff
@@ -109,65 +109,39 @@ public defn connect-gnd-pins ():
...
-          match(voltage: Double) :
-            if is-resistor?(inst) :
-              property(inst.operating-point) = resistor-operating-point(voltage, resistance!(inst))
-            if is-capacitor?(inst) :
-              property(inst.operating-point) = capacitor-operating-point(voltage)
-    else :
-      gnd-type-error!(self, gnd) when gnd is-not False
+    for i in component-instances(self) do :
+      if has-property?(i.resistance) or has-property?(i.capacitance) or has-property?(i.inductance):
+        if length(pins(i)) == 2 :
```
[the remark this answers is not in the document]
bhusang: Yea, some resistors will have 3 pins -- 2 regular pins and one for thermal or shield.
https://www.vishay.com/docs/30201/rhnh.pdf

utils/generator-utils.stanza
```diff
@@ -221,6 +195,23 @@ public defn set-power-source (pos:JITXObject, neg:JITXObject, v-in:Double):
...
+      for n in nets(item) do :
+        if has-property?(n.net-voltage) :      
+          add(voltages, property(n.net-voltage))
+      ;If at least one net has a voltage property.
+      if not empty?(voltages) :
+            
+        ;Now assign the voltage to all of the component pins in 
+        ;the module.
+        for p in component-pins(item) do :
+          property(p.net-voltage) = voltages[0]
```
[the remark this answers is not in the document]
bhusang: Should we also check that a 0.0v net-voltage has been assigned? Like for gnd?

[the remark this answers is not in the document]
bhusang: And what if the first voltage in the Vector is 0.0v from gnd?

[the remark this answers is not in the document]
bhusang: The bigger question is that when you are assigning the voltages[0] to the pins of the component attached to that net, are those *all* of the pins of that component? Or just the connected pins? And do you ignore non-populated components or NC pins?

utils/generator-utils.stanza
d-haldane: You can have negative voltages

modules/power-regulators.stanza
d-haldane: That could be assigned here, yes

utils/generator-utils.stanza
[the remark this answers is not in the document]
d-haldane: Now that tolerance is equalable, we should add a check that all assigned voltages on a set of connected items are identical

utils/generator-utils.stanza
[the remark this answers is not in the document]
d-haldane: Check for 2 pins sounds good

utils/generator-utils.stanza
```diff
@@ -221,6 +195,23 @@ public defn set-power-source (pos:JITXObject, neg:JITXObject, v-in:Double):
...
 
+public defn propagate-net-voltages () :
+  inside pcb-module:
+    for item in all-connected-items(self) do :
+      
+      ;First gather all the voltages on all connected nets.
+      val voltages = Vector<Toleranced>()
+      for n in nets(item) do :
+        if has-property?(n.net-voltage) :      
+          add(voltages, property(n.net-voltage))
```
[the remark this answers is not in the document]
bhusang: changed

utils/generator-utils.stanza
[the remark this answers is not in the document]
bhusang: Will need assistance on this.. not changing for now.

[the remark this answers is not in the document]
d-haldane: I'll do that real quick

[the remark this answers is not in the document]
d-haldane: Inside `items`, everything is electrically connected. I added a check to make sure you didn't set two different `net-voltages` inside that set. It is not checking across all nets in the design here - just each set of connected nets.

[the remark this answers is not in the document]
bhusang: I looked again at @CuppoJava's doc and it was clear how the code is structured. I copied more of his example to improve the error messages.

utils/generator-utils.stanza
```diff
@@ -206,6 +206,9 @@ public defn propagate-net-voltages () :
           add(voltages, property(n.net-voltage))
       ;If at least one net has a voltage property.
       if not empty?(voltages) :
+        for i in 1 to length(nets(item)) do :
+          if voltages[i] != voltages[0] :
+            fatal("Mismatched voltage")
```
PhilippeFerreiraDeSousa: You would do:
```
if length(unique(voltages)) != 1 : 
  fatal("Mismatched voltages %_ on nets %_" % [voltages, map(ref, nets(item))])
```

utils/generator-utils.stanza
```diff
@@ -221,6 +209,38 @@ public defn set-power-source (pos:JITXObject, neg:JITXObject, v-in:Double):
...
+          match(ref?(net)) :
+            (f:False) : add(net-names, to-string("unnamed-net-%_" % [c]))
+            (r:Ref) : add(net-names, r)
+
+        ;Let's check that all the voltages happen to be the same.
+        ;If they are different, then something has gone wrong.
+        if length(voltages) > 1 :
+          for i in 1 to length(nets(item)) do :
+            if voltages[i] != voltages[0] :
+              fatal("Mismatched net-voltage on net(s) %," % [net-names])
```
PhilippeFerreiraDeSousa, line 237: You would do:
```
if length(unique(voltages)) != 1 : 
  fatal("Mismatched voltages %_ on nets %_" % [voltages, map(ref, nets(item))])
```

utils/generator-utils.stanza
```diff
@@ -221,6 +209,38 @@ public defn set-power-source (pos:JITXObject, neg:JITXObject, v-in:Double):
...
+
+        ;Now assign the voltage to all of the component pins in 
+        ;the module.
+        for p in component-pins(item) do :
+          property(p.net-voltage) = voltages[0]
```
PhilippeFerreiraDeSousa, line 242: The original code was
```
if not empty?(voltages) : 
  for p in component-pins(item) do :
  property(p.net-voltage) = voltages[0]
```
The new version will crash on `voltages[0]` when voltages is empty, which is very often the case.

PhilippeFerreiraDeSousa: This version is good:
```
        if not empty?(voltages) :
          if length(unique(voltages)) != 1 : 
            fatal("Mismatched voltages %_ on nets %_" % [voltages, net-names])
          ;Now assign the voltage to all of the component pins in the module.
          for p in component-pins(item) do :
            property(p.net-voltage) = voltages[0]

PhilippeFerreiraDeSousa: Ok I just saw that you had put a `if not empty?(voltages) :` at the very beginning, I missed it

utils/generator-utils.stanza
bhusang: Nicely succinct!

PhilippeFerreiraDeSousa (approved):

PR #254, opened 2022-07-22
Add REF and VALUE to spst

ariel-hi:
This PR adds a REF text and VALUE text to `pcb-symbol spst` in `utils/symbols.stanza`.

I'd like to learn more about the situational usage of such text, so I have some questions for reviewers:
- `altium-gost-gnd-...` earth and power symbols have neither REF nor VALUE. I see other Altium net symbols have a VALUE -- should a VALUE be added to these two symbols?
- The testpoint and hole symbols do not have VALUEs. That makes because of how generic those concepts are.
- And lastly, should `spst` have a REF and VALUE? I'm effectively asking if this PR should go through, and if the symbol was designed to intentionally not have the texts.

### DIFF
--- a/utils/symbols.stanza
+++ b/utils/symbols.stanza
@@ -814,6 +814,9 @@ public pcb-symbol spst (n-pside:Int, n-tside:Int) :
       name-size = 0.762
   unit-line([[( - xpoff), (- 1.0)], [xpoff, (- 1.0)]])
 
+  unit-ref([1.0, 2.0])
+  unit-val([1.0, 1.0])
+
 ; TODO: Relay symbol parametric in connections
 
 ; ====== RF ====================================================================


d-haldane (approved):

d-haldane:
It certainly should - the omission of these is very rarely intentional

ariel-hi:
@d-haldane Good to know! I will change only `spst` here. There are bigger items to get to.

bhusang:
- altium-gost-gnd-... earth and power symbols have neither REF nor VALUE. I see other Altium net symbols have a VALUE -- should a VALUE be added to these two symbols?
   No, power and ground symbols may have a implicit REF but there is no obvious reason that they need a VALUE

The testpoint and hole symbols do not have VALUEs. That makes because of how generic those concepts are.
  Agreed that they generally don't have values, however, I have sometimes added a string value to a testpoint which allows me to use that string mapped to the silkscreen layer to have a descriptive message on the test point. As long as we still permit this, we should be ok.

And lastly, should spst have a REF and VALUE? I'm effectively asking if this PR should go through, and if the symbol was designed to intentionally not have the texts.

Not sure about the VALUE because I am not sure what that would map to. The REF is useful.

bhusang (approved):

ariel-hi:
Thanks @d-haldane and @bhusang

These utility testpoints won't have a value text. It is permitted in the sense that you can define your own testpoint with a value.

PR #256, opened 2022-07-22
Speed up component check runtime

PhilippeFerreiraDeSousa:
BLE-Mote
image

### DIFF
--- a/utils/checks.stanza
+++ b/utils/checks.stanza
@@ -254,7 +254,7 @@ pcb-check no-connect-pins-not-connected (module:JITXObject) :
     val populated-connected-pins = populated-connected-pins(p)
 
     #CHECK(
-      condition =            not has-populated-connected-pins?(p)
+      condition =            empty?(populated-connected-pins)
       name =                 "IO pin checks"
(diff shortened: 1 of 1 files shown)


utils/checks.stanza
```diff
@@ -249,7 +254,7 @@ pcb-check no-connect-pins-not-connected (module:JITXObject) :
     val populated-connected-pins = populated-connected-pins(p)
```
ariel-hi, line 254: You may want to delete this line!

ariel-hi: Hmm, I see that this is used in the locators. That's ok then!
In this case it is not necessary to call `has-populated-connected-pins?` because you have access to the pins already.

ariel-hi (commented):
Good speed up!
I would suggest undoing the change on line 257

utils/checks.stanza
PhilippeFerreiraDeSousa: I reverted the line

ariel-hi (approved):

bhusang (approved):

PR #258, opened 2022-07-23
added md doc

bhusang:
Added table to explain sample designs in the ocdb library


ariel-hi (approved):
Great to have a file summarizing our example designs!

PR #259, opened 2022-07-25
Speed up DRC landpattern checks

PhilippeFerreiraDeSousa:
`component-instances(module)` is already a recursive function that returns all components under module, whether direct or indirect children.

PR goes with https://github.com/JITx-Inc/jitx-client/pull/1293.

I need to benchmark the speed up tomorrow to confirm uncommenting the check is acceptable.

### DIFF
--- a/utils/checks.stanza
+++ b/utils/checks.stanza
@@ -1734,16 +1734,17 @@ public pcb-check check-min-solder-mask-sliver (lp:LandPattern, shapes:Seqable<Sh
       locators =             [lp]
     )
 
-; Runs all the land pattern checks on either an instance or definition. 
-public defn check-landpattern (component:JITXObject|JITXDef) :
+; Runs all the land pattern checks on either an instance or definition.
+; Not currently used in the design checks
+public defn check-landpattern (obj:JITXObject|JITXDef) :
   ; extract the land pattern
-  val lp = 
-    match(component) :
+  val lp =
+    match(obj) :
       (landpattern:LandPattern) :
         landpattern
       (d:JITXDef) :
         landpattern(d)
-      (i:JITXObject) : 
+      (i:JITXObject) :
         landpattern(instance-definition(i))
 
   ; perform the checks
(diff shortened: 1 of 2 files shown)


utils/checks.stanza
```diff
@@ -1764,20 +1762,14 @@ public defn check-landpattern (component:JITXObject|JITXDef) :
 ; Runs landpattern checks recursively on all the lands in a module
 public defn check-all-landpatterns (module:JITXObject) : 
   inside pcb-module :
-    val components = to-tuple(component-instances(module))
-    val modules    = filter({not contains?(components, _)}, instances(module))
-    for component in components do :
-      check-landpattern(component)
-    for module in modules do :
-      check-all-landpatterns(module)
+    do(check-landpattern, component-instances(module))
```
PhilippeFerreiraDeSousa, line 1765: Better but far from optimum. If there is 100x the same LED or resistor, we don't want 100 times the same check.  We should group them by component definition. @ariel-hi do you know the introspection method to do that?

m-hilgendorf (approved):

utils/checks.stanza
```diff
@@ -1755,29 +1754,54 @@ public defn check-landpattern (component:JITXObject|JITXDef) :
       check check-min-copper-width(lp, side)
       check check-min-solder-mask-sliver(lp, layer(lp, SolderMask(side)))
 
-    if has-property?(component.bga) and property(component.bga):
+    if property?(obj.bga, false):
       check check-bga-pitch(lp)
 
-    if has-property?(component.leaded) and property(component.leaded):
+    if property?(obj.leaded, false):
```
ariel-hi, line 1760: I see you using the new macro 😄

utils/checks.stanza
```diff
@@ -1755,29 +1754,54 @@ public defn check-landpattern (component:JITXObject|JITXDef) :
...
+    for side in [Top, Bottom] do :
+      check check-min-copper-copper-space(lp, side)
+      check check-min-copper-width(lp, side)
+      check check-min-solder-mask-sliver(lp, layer(lp, SolderMask(side)))
+
+    if bga(info) :
+      check check-bga-pitch(lp)
+
+    if leaded(info) :
       check check-leaded-pitch(lp)
```
ariel-hi, line 1785: Consider merging this logic with the duplicate logic in `check-landpattern`

utils/checks.stanza
PhilippeFerreiraDeSousa: You got me!

PR #261, opened 2022-07-25
added stanza.proj to make design directory usable

bhusang:


designs/stanza.proj
```diff
@@ -0,0 +1 @@
+include "../stanza.proj"
```
PhilippeFerreiraDeSousa, line 1: Is the problem for developers or users.

If it is for developers:

What you are missing is to export
```
export STANZA_CONFIG=$JITX_ROOT
```
This will allow you to refer to the packages in the repo from anywhere in your computer. I added it to the new hire doc yesterday.

Otherwise, it is open-components-database/stanza.proj that should include `designs/stanza.proj` so that you can `import ocdb/ble-mote` from a repl anywhere.

bhusang: So if a user clones our ocdb and attempts to run a design in the `designs` directory, obviously it will not run out of the box. And given that we have no instructions on how to do so, this additional file allows a new user to directly run designs in the design directory. That was the only case that I was trying to solve -- user access ocdb and wants to run a design without reading any instructions.

bhusang: Granted that we would want them to read instructions but, hey, hardware users are generally instruction-disinclined

PhilippeFerreiraDeSousa: Making sure, for users yes.

PhilippeFerreiraDeSousa (approved):

PR #262, opened 2022-07-26
removed extraneous equal method and fixed check message

bhusang:


PhilippeFerreiraDeSousa (approved):
Good

utils/property-structs.stanza
```diff
@@ -143,9 +143,6 @@ doc:"A well-defined state for a device's power, for example hibernation. \
 public pcb-struct ocdb/property-structs/PowerState <: Equalable&Hashable : 
   name:String,
   powered?:True|False
-defmethod equal? (l:PowerState, r:PowerState) : 
-  name(l)     == name(r) and 
-  powered?(l) == powered?(r)
 defmethod hash (p:PowerState) :
```
PhilippeFerreiraDeSousa, line 146: You also should remove the hash function

PhilippeFerreiraDeSousa: I added
```
    with :
      writer => true
      equalable => true
      hashable => true
```
to pcb-structs

PhilippeFerreiraDeSousa: You also do not need to make pcb-struct inherit from `Equalable&Hashable` explicitly because the macro already makes them inherit from this

bhusang: Done.

PhilippeFerreiraDeSousa (approved):
Awesome

PR #263, opened 2022-07-28
Update CP2105 circuit, remove view calls to component, update design template

d-haldane:


utils/checks.stanza
```diff
@@ -747,16 +747,6 @@ public pcb-check generic-pin-check (p:JITXObject):
...
-    #CHECK(
-      condition =            has-property?(p.net-voltage),
-      name =                 "Generic pin checks"
-      description =          "Check net-voltage property on generic pin"
-      category =             "Pin Checks"
-      subcheck-description = "Verify the generic pin has a net-voltage property attached",
-      pass-message =         "Pin %_ has the net-voltage property (%_V) attached" % [ref(p), property(p.net-voltage)],
-      info-message =         "Pin %_ does not have the net-voltage property attached" % [ref(p)],
-      locators =             [p]
-    )
```
bhusang, line 759: What prompted the removal of this info check?

d-haldane: It triggers on every pin with the generic pin property. `net-voltage` was meant for power rails so this was triggering a bunch of info checks that did not point towards improving the design

bhusang: Oh -- I assumed net-voltage was for more than just power nets (i.e. any net where the user wanted to check the voltage limit).

d-haldane: Maybe we could bring it back, but make sure it's not trigger on unconnected pins?

bhusang: Well let's leave it off for now given the divergence in usage. Unconnected pins (with the no-connect property) were excluded from pin checks at one point. Not sure what the current status of property(pin.no-connect) is. For sure the checks.stanza looks for the no-connect property 
```
; Checks if a pin is marked "no-connect"
defn no-connect? (p:JITXObject) -> True|False : 
  has-property?(p.no-connect) or
  has-property?(p.nc) or 
  has-property?(p.NC)
```

components/si-labs/CP2105.stanza
```diff
@@ -78,9 +78,8 @@ public pcb-component component :
   assign-landpattern(qfn-landpattern(0.5, 4.0, 24, 0.25, 0.4, [2.7 2.7]))
 
   property(REGIN.power-pin) = PowerPin(min-max(3.0, 5.25))
-  property(VDD.power-pin) = PowerPin(min-max(3.0, 3.6))
-  eval-when has-property?(VDD.net-voltage) :
-    property(VIO.power-pin) = PowerPin(min-max(1.8, max-value(property(VDD.net-voltage))))
+  eval-when has-property?(REGIN.net-voltage) :
+    property(VIO.power-pin) = PowerPin(min-max(1.8, max-value(property(REGIN.net-voltage))))
   eval-when has-property?(VIO.net-voltage) :
```
bhusang, line 83: Given that VDD does not need to be tied to VIO because the IO can be operated at a lower voltage so VIO is treated as a power supply input for the CP2105. External devices can be powered from VDD, you might need to support VDD as a supply output as as follows:
```
  ; we assume that REGIN voltage is always copied to VDD (and vice-versa???)
  eval-when has-property?(REGIN.net-voltage) :
    if property(REGIN.net-voltage) > 3.0 : ; this is likely not spec'ed correctly in the datasheet
      property(VDD.power-pin) = PowerPin(min-typ-max(3.3, 3.45, 3.6))
```

bhusang (dismissed):

components/si-labs/CP2105.stanza
```diff
@@ -99,10 +105,11 @@ public pcb-module module :
   pin gnd
   
   public inst xcvr : ocdb/si-labs/CP2105/component
+  ; connected as bus-powered device
   net (usb-2.vbus.vdd xcvr.VBUS xcvr.REGIN)
   cap-strap(xcvr.REGIN, gnd, 1.0e-6)
   net (usb-2.vbus.gnd xcvr.GND xcvr.EPAD gnd)
-  net (xcvr.VIO vio)
+  net (xcvr.VIO xcvr.VDD vio) ; connect VDD and VIO together
```
d-haldane, line 112: @bhusang I don't think we want to hard-wire vio to vdd - the separate io voltage is a common use case (including being used on ble-mote).

You can just expose `vdd` and `vio` and let the user manage it

bhusang: @d-haldane  That's a better solution but I was thinking to preserve the connectivity that this component was already using. If that's not the case, I will separate vdd out.

d-haldane: Better to make sure the use case is supported - port interfaces are pretty easy to update

bhusang (approved):

PR #264, opened 2022-07-29
Added USB-C bundle type, termination function, enum, and cleaned up pin names

electricfungi:


modules/terminations.stanza
```diff
@@ -28,6 +28,43 @@ public pcb-enum ocdb/terminations/LVDSTerminationMethod :
...
   LVDSTerminationACDualResistorSplitReduced
 
+public pcb-enum ocdb/terminations/USB-C-ConfigurationChannel :
+  SourceDefault
+  Source5V-1A5
+  Source5V3A0
+  SinkDefault
+
+; Terminates USB-C based on pull voltage and USB-C configuration
+public defn terminate-usb-c (usb-c-bundle:JITXObject, pull-up-net:JITXObject, pull-up-voltage:Double, method:USB-C-ConfigurationChannel) :
```
bhusang, line 38: We should move away from using Doubles as voltages and replace them with Toleranced. So this might convert to:
```
pull-up-voltage:Toleranced
```
and the invocation could be something like `typ(5.0v)`

modules/terminations.stanza
```diff
@@ -28,6 +28,43 @@ public pcb-enum ocdb/terminations/LVDSTerminationMethod :
...
+        public inst term : chip-resistor(termination-resistor-value)[2]
+        net (usb-c-bundle.vbus.vdd term[0].p[2] term[1].p[2])
+        net (term[0].p[1] usb-c-bundle.cc[1])
+        net (term[1].p[1] usb-c-bundle.cc[2])
+      SinkDefault:
+        public inst term : chip-resistor(termination-resistor-value)[2]
+        net (usb-c-bundle.vbus.gnd term[0].p[2] term[1].p[2])
+        net (term[0].p[1] usb-c-bundle.cc[1])
+        net (term[1].p[1] usb-c-bundle.cc[2])
+
```
bhusang, line 67: In your equation for the term resistor value, do you have a guarantee that pull-up-voltage/[80e-6, 180e-6, 330e-6] is always greater than 5.1e3? Otherwise the argument to chip-resistor() can be negative.

bhusang: In addition, you should refactor your logic to remove repeated statements:
```
  inside pcb-module:
    val term-current = 
      switch(method) :
        SourceDefault:
          80.0e-6
        Source5V-1A5:
          180.0e-6
        Source5V3A0:
          330.0e-6
        SinkDefault:
          0.0

    val termination-resistor-value = 
      if term-current != 0.0 :
        closest-std-val((pull-up-voltage / term-current) - 5.1e3, 5.0)
      else :
        5.1e3 ; Default Sink termination value
    public inst term : chip-resistor(termination-resistor-value)[2]
    net (usb-c-bundle.vbus.vdd term[0].p[2] term[1].p[2])
    net (term[0].p[1] usb-c-bundle.cc[1])
    net (term[1].p[1] usb-c-bundle.cc[2])
```

bhusang: If you use pull-up-voltage as Toleranced, then the access to that should be (depending on the limit analysis):
```
typ(pull-up-voltage) 
or 
min(pull-up-voltage)
or
max(pull-up-voltage)
```

bhusang (changes requested):
Left some comments on improvements.

modules/terminations.stanza
electricfungi: the netting is different. I would have to add two sets of switch statements to do this

modules/terminations.stanza
electricfungi: fixed

modules/terminations.stanza
bhusang: I think that you can do it with one switch statement returning the val of current and net as a tuple. I missed the gnd net used in the SinkDefault, but the same structure can be used.

d-haldane:
You added `closest-std-val-range` and made `union-std-vals` public but didn't use them anywhere in this PR or add any tests using them. Whence these changes?

The CC logic for USB C doesn't belong in terminations - it's not a termination like the LVDS or CAN functions in this file.

The arguments to the 'termination' function are confusing to me.
```
 public defn terminate-usb-c (usb-c-bundle:JITXObject, pull-up-net:JITXObject, pull-up-voltage:Double, method:USB-C-ConfigurationChannel) : 
```

- You require a `pull-up-net` and a `pull-up-voltage` but resistors can be pull down, making these arguments confusing. And you have to supply a voltage > 0.41 volts even if it's not used.

- The Double|Toleranced argument should just be Toleranced

- The pull-down net is inferred, which makes it extra confusing to require two separate pull up arguments.

- You require a net *and* its voltage. It would be better to rely on the `net-voltage` annotation and guard execution with an eval-when.

- The elements of `USB-C-ConfigurationChannel` have names like source and sink, but the standard seems to have names like Upstream facing port and downstream facing port. These names in the enum should match the standard to avoid confusion

d-haldane:
For names, referring to https://ww1.microchip.com/downloads/en/appnotes/00001953a.pdf

> • Downstream Facing Port (DFP) - A host or downstream hub port. Typical of a legacy standard Type-A port.
> • Upstream Facing Port (UFP) - A device or upstream hub port. Typical of a legacy standard Type-B port.
> • Dual-Role Port (DRP) - A port that transitions between DFP and UFP port states until an attach event occurs.
> DRPs may be dynamically swapped using USB Power Delivery Protocol Negotiation after an initial attach event.
> • Power Source or Provider - A source of 5V-20V up to 5A. Typical of a legacy standard Type-A port.
> • Power Sink or Consumer - A sink of 5V-20V up to 5A. Typical of a legacy standard Type-B port.

When you're implementing a general function like this, the functionality should covert he common configurations in a legible way.

electricfungi:
I'm using source and sink per USB-C standard page 33 https://www.usb.org/sites/default/files/USB%20Type-C%20Spec%20R2.0%20-%20August%202019.pdf Power delivery applications will use dedicated chips, not the resistors this function adds. We can always expand the enum if the use case appears where power delivery enums are needed. I think I addressed other concerns

utils/generic-components.stanza
```diff
@@ -654,22 +653,34 @@ public pcb-module micro-usb-connector :
...
 
   net D- (usb-2.data.N con.DN1, con.DN2)
   net D+ (usb-2.data.P con.DP1, con.DP2)
-
+  net (usb-c.data[1] usb-2.data)
   ; RC filter connection for chassis gnd
   res-strap(con.SHIELD, con.GND, 100.0e3)
   cap-strap(con.SHIELD, con.GND, 0.1e-6)
-
+  net (con.SBU1 usb-c.sbu[1])
```
d-haldane, line 669: These connections would be better moved to the belfuse definition, to be consistent with the hroparts connector. Create the local module in that package to map to usb-c, and then either connector could be using in this reusable module.

electricfungi: fixed

utils/generic-components.stanza
```diff
@@ -652,25 +651,27 @@ public pcb-module micro-usb-connector :
...
-  cap-strap(con.SHIELD, con.GND, 0.1e-6)
-
-  ; 5V Downstream USBC device configuration
-  res-strap(con.CC1, con.GND, 5.1e3)
-  res-strap(con.CC2, con.GND, 5.1e3)
-
-  property(con.vbus) = PowerSupplyPin(5.0, 0.5)
+public defn  usb2-on-a-usb-c-connector ():
+  usb2-on-a-usb-c-connector(Belfuse-SS-52400-003)
+public pcb-module usb2-on-a-usb-c-connector (conn-mpn:ocdb/generic-components/usb-c-connector):
```
d-haldane, line 660: I like the idea of providing the connector as an argument to this module. 

You can avoid the enum + switch, by directly providing the instantiable:
```
 public pcb-module usb2-on-a-usb-c-connector (con-inst:Instantiable) :
  inst con : con-inst
```
Called with 
```
usb2-on-a-usb-c-connector(ocdb/belfuse/SS-52400-003/module)
```

I would drop the `usb2` definitions in the component defpackages, and use this module to map `usb2` by instantiating the connector and using the configuration function.

d-haldane:
@electricfungi This looks ready for final cleanup! I like the reuse you enabled for configuration, and using usb-c for usb-2 with anything supporting the right pins.

Please fix:

- The unused ` public pcb-enum ocdb/generic-components/usb-c-connector:` 
- The merge conflicts in generic-components with this branch. (Always be merging dev into your branch when you have an active PR to avoid these issues.)

d-haldane (approved):

PR #265, opened 2022-08-03
Adding parts from led-display project to OCDB

bhusang:
For ease of use, adding parts to ocdb to facilitate usage of project by all


m-hilgendorf (approved):
Design verified components, big win!

PR #268, opened 2022-08-08
Minor cleanup of design template

electricfungi:


bhusang (approved):
Seems like this was a spurious commit to begin with?

PR #270, opened 2022-08-12
Updates to ble-mote

d-haldane:
Improving the quality of the implementation:
- Sensors factored out to separate module
- System generator that takes processor and outline as arguments
- Dual exports to two designs


components/artwork/board-text.stanza
```diff
@@ -51,6 +51,7 @@ public pcb-component text (title:String, size:Double, outline-size:Double) :
   val sym = generic-symbol
   symbol = sym()
   reference-prefix = "TXT"
+  property(self.rated-temperature) = false
```
bhusang, line 54: Interesting -- does it make sense to even add this property to artwork simply because the check would be triggered otherwise? We may want to investigate how to make the check more specific. For now it's ok.

components/espressif/esp32-wroom-32.stanza
```diff
@@ -124,11 +124,12 @@ public pcb-module module (pn:esp32-wroom-32mpn, add-serial-programming-circuit?:
   port power : power
   pin boot
   pin en
+  pin gnd
```
bhusang, line 127: Why add a pin for gnd when that it should be incorporated in the power port?

components/si-labs/CP2105.stanza
```diff
@@ -20,32 +20,6 @@ public pcb-component component :
...
-  pin CTS_SCI
-  pin RTS_SCI
-  pin RXD_SCI
-  pin TXD_SCI
-  pin DSR_SCI
-  pin DTR_SCI
-  pin DCD_SCI
-  pin EPAD
-
   pin-properties :
```
bhusang, line 23: @ariel-hi Just added a Linear task to take all pin:Ref's in a pin-properties: block to be automatically promoted to the component level as well (making the explicit definition redundant). This is ok until that PR is done.

designs/ble-mote.stanza
```diff
@@ -18,98 +18,92 @@ defpackage ocdb/designs/ble-mote :
...
+pcb-bundle pms7003-interface :
   port vin : power
   port vin-mcu : power
   pin reset
   pin set
   port uart : uart()
+
+pcb-module pms7003 :
+  port interface : pms7003-interface
+
```
bhusang, line 37: Interesting -- create an interface to a module as a bundle which makes it more efficient and concise to access. I like this approach.

designs/ble-mote.stanza
```diff
@@ -18,98 +18,92 @@ defpackage ocdb/designs/ble-mote :
...
 
-  ; Add a Bluetooth SoC
-  public inst proc : ocdb/espressif/ESP32-PICO-D4/module
-  public inst ant : inverted-f-antenna
-  net (proc.ant, ant.rf-in)
-  net (ant.gnd gnd)
-  place(ant) at loc(9.0, 5.0, -90.0) on Top
+  property(VBUS.net-voltage) = min-max(4.8, 5.2)
+  property(P5V0.net-voltage) = min-max(4.9, 5.1)
+  property(P3V3.net-voltage) = typ(3.3)
```
bhusang, line 81: I might suggest `min-typ-max(2.97, 3.30, 3.63)` for the specification of P3V3. Not a big deal, but it's more attuned to what users might need to check against.

designs/ble-mote.stanza
```diff
@@ -118,77 +112,143 @@ pcb-module sensor-mote :
...
-  net D- (usb.usb-2.data.N)
-  net D+ (usb.usb-2.data.P)
-
-  property(VBUS.net-voltage) = min-max(4.8, 5.2)
-  property(P5V0.net-voltage) = min-max(4.9, 5.1)
-  property(P3V3.net-voltage) = typ(3.3)
-  property(gnd.net-voltage) = typ(0.0)
-
   ; Set up netclasses
   val power-net-class = NetClass(`Power, [`min-trace => 0.50])
```
bhusang, line 128: Learn something new everyday -- a property called `power-net-class` with a `NetClass()` assignor. But you use symbols (``) in the `NetClass()` function?

designs/ble-mote.stanza
```diff
@@ -118,77 +112,143 @@ pcb-module sensor-mote :
...
+; Run the design checks
+; =====================
+defn check-design () :
+  run-design(true)
+
+; ===============================================
+; Export two versions of the sensor-system to CAD
+; ===============================================
+defn export-designs () :
+  set-export-backend(`altium)
```
bhusang, line 218: Could we pass an argument into this function to select which board version to export? That way the common code can be minimized. The overhead is that you then have to call the `export-designs()` function twice from somewhere else? I assume it's the REPL?

bhusang (dismissed):

components/artwork/board-text.stanza
d-haldane: That's right - changing the convention that geometry 'components' should be skipped in rated temperature checks.

components/espressif/esp32-wroom-32.stanza
d-haldane: Just a convenience - many things attach to ground and I get tired of typing power-3v3.gnd

designs/ble-mote.stanza
d-haldane: Yes! It works well for systems comprised by multiple boards as well - the bundles stack together really nicely

designs/ble-mote.stanza
d-haldane: Yes - I'll make a note to add docs for this.

designs/ble-mote.stanza
d-haldane: Done.

bhusang:
This PR needs to be rebased on the coming ocdb with the new namespace.

bhusang (approved):
Looks good

m-hilgendorf:
This PR needs to be reverted or fixed, it introduces conflicting `stanza.proj` files with the wrong naming convention (the new keystone components). 

After #271 was merged, new `stanza.proj` files should be forbidden in sub-directories

m-hilgendorf:
Working on it! Just need to delete the stanza.proj file.

PR #271, opened 2022-08-19
Unify Component .proj files.

m-hilgendorf:


bhusang (approved):
Not sure how this makes including packages at the global level any easier, but your changes all look good.

m-hilgendorf:
@bhusang this is required for the coming changes to .proj files in stanza 0.17.16. The prefix is now required.

PR #275, opened 2022-09-07
Add an example of a NTC thermistor

bhusang:


components/murata/NCP03WF104F05RL.stanza
```diff
@@ -0,0 +1,55 @@
...
+  port p : pin[[1 2]]
+
+  val sym = resistor-sym(ResistorTherm)
+  symbol = sym(p[1] => sym.p[1], p[2] => sym.p[2])
+  val pkg = ipc-two-pin-landpattern("0201")
+  landpattern = pkg(p[1] => pkg.p[1], p[2] => pkg.p[2])
+
+  val generic-props = GenericPin(min-max(0.0, 5.0), 2000.0)
+
+  property(p[1].generic-pin) = generic-props
```
d-haldane, line 33: generic-pin was meant more for IC pins. I would replace this with a property associated with resistors, like 

```
rated-voltage = 5.0
```

bhusang: Ok, that makes sense. I just looked for any "compatible" property that had a max-voltage associated with it.

bhusang: BTW, the PWL macro only takes tuples of doubles, but we may want to upgrade it to accept a Tuple<Double, Toleranced> -> Toleranced? (another PR, I think)

bhusang: Technically the `rated-voltage` property should be a relative voltage (i.e. the max voltage difference between p[1] and p[2] for rated-voltage = 5.0. But I am not sure how to code that (yet).

d-haldane (approved):

PR #276, opened 2022-09-15
Refactor Passive Component Checks

m-hilgendorf:


utils/passive-checks/capacitor-checks.stanza
```diff
@@ -0,0 +1,315 @@
+#use-added-syntax(jitx)
```
m-hilgendorf, line 1: Keep around the old code for posterity

utils/checks.stanza
```diff
@@ -54,14 +57,15 @@ public defn erc-check-design (module:JITXObject):
   inside pcb-module:
     for i in populated-components(module) do :
       check rated-temperature(i)
-      check capacitor-component(i) when has-property?(i.capacitor)
-      check resistor-component(i)  when has-property?(i.resistor)
-      check inductor-component(i)  when has-property?(i.inductor)
-      check diode-component(i)     when has-property?(i.diode)
```
m-hilgendorf, line 60: This is not an accident: the diode component checks are redundant (only pin connectivity is checked, which is handled elsewhere)

utils/checks.stanza
```diff
@@ -54,14 +57,15 @@ public defn erc-check-design (module:JITXObject):
...
+      
+      check-capacitor(i) when has-property?(i.capacitor)
+      check-resistor(i)  when has-property?(i.resistor)
+      check-inductor(i)  when has-property?(i.inductor)
```
m-hilgendorf, line 63: Implementation moved to `passive-checks/*`

utils/passive-checks/capacitor-checks.stanza
```diff
@@ -0,0 +1,315 @@
...
+  import ocdb/utils/design-vars
+  import ocdb/space-derating
+  import ocdb/utils/generator-utils
+  import ocdb/utils/property-structs
+  import ocdb/utils/tolerance
+  import ocdb/utils/passive-checks/utils
+
+;==============================================================================
+;========================== Capacitor Checks ==================================
+;==============================================================================
```
m-hilgendorf, line 16: This is the best example of the new checks architecture.

utils/checks.stanza
bhusang: We should add 1-2 more diode checks (like reverse breakdown voltage, max forward current) but your simplification is ok. Let's plan to add diodes to the passive-checks.

bhusang (approved):
The rewrite looks great. I like how you have repartitioned the checks into separate files with a clear distinction on finding the properties needed. Going forward this will be a great base from which to evolve the checks comprehensively.

derating/space-derating.stanza
```diff
@@ -5,31 +5,31 @@ defpackage ocdb/space-derating :
 public var DESIGN-TEMPERATURE:Double = 0.0
 
 ; RESISTOR DERATING PARAMETERS
-public val DERATE-RESISTOR-POWER = 0.6
-public val DERATE-INDUCTOR-POWER = 0.6
+public var DERATE-RESISTOR-POWER = 0.6
+public var DERATE-INDUCTOR-POWER = 0.6
```
m-hilgendorf, line 9: SIdenote: these are now `var` in case a user wishes to customize deratings in a design.

PR #278, opened 2022-09-16
Removed old checks and refactored pin checks

m-hilgendorf:
Done : 
- Comprehensive pin connectivity checks 
- Move checks into separate file structure

Two things to solve  :

- The context for a pin is hard to lookup without new introspections. I'd like it to be <ref-des>.<pin-name>, but that isn't possible at the moment - so I'm setting for keeping the ref with a plan to update it. 
- We don't have a notion of checking pins that don't have properties - ideally we would warn against using pins without any properties that can't be verified.


utils/checks.stanza
```diff
@@ -728,335 +725,10 @@ public pcb-check connected (p:JITXObject|Tuple<JITXObject>) :
...
-    description =          "Check reset pin properties"
-    category =             "Pin Checks"
-    subcheck-description = "Verify the reset pin is connected",
-    pass-message =         "Reset pin %_ is connected" % [ref(p)],
-    fail-message =         "Reset pin %_ is not connected" % [ref(p)],
-    locators =             [p]
-  )
-
-
-defn check-rated-voltage (p:JITXObject) :
```
m-hilgendorf, line 839: This was not used anywhere and the properties used are not up to date

utils/checks.stanza
```diff
@@ -728,335 +725,10 @@ public pcb-check connected (p:JITXObject|Tuple<JITXObject>) :
...
-        category =             "rated-voltage"
-        subcheck-description = "Check the pin's rated-voltage property is within the design voltage specification",
-        pass-message =         "Pin %_ rated voltage %_V low is correctly within the design supply specification %_V" % [ref(p),property(p.rated-voltage)[0],v[0]],
-        fail-message =         "Pin %_ rated voltage %_V low is incorrectly outside the design supply specification %_V" % [ref(p),property(p.rated-voltage)[0],v[0]],
-        locators =             [p]
-      )
-    (r, v) : false
-
-; Checks if voltage on a pin is within acceptable limits
-public pcb-check voltage-levels (p:JITXObject, range:[Double,Double,Double]) :
```
m-hilgendorf, line 898: Same as above

bhusang (approved):
Looks good. Same good idea to further separate the checks into separate files.

PR #282, opened 2022-09-20
Fix the soldermask around the BGA pads for the Nordic part

bhusang:


components/nordic/nRF52840.stanza
```diff
@@ -19,80 +19,86 @@ defpackage ocdb/components/nordic/nRF52840 :
   import ocdb/utils/checks
   import ocdb/utils/passive-checks/resonator-checks
 
+defn nrf-bga-pad (copper-diameter:Double) :
```
m-hilgendorf, line 22: Can we use the BGA landpattern generator for this part?

m-hilgendorf: Oh yea that's a weird one - it might be possible through some custom pad-omission hacking but your approach is easier atm.

m-hilgendorf (approved):

PR #283, opened 2022-09-20
Voltage Divider Toleranced Arithmetic

ariel-hi:
Cleaning up voltage divider calculations using upcoming Toleranced arithmetic.

I couldn't consolidate all the way because high resistance and low resistance work oppositely. Please supply recommendations!


modules/solvers/voltage-divider.stanza
```diff
@@ -91,46 +97,61 @@ defn get-resistance-values (goal-r: Double, r-tol: Double, current-tol: Double)
...
+  val vo-nom = (v-in-typ * r-lo-typ / (r-lo-typ + r-hi-val-typ)) where :
+    val v-in-typ = typ-value(v-in)
+    val r-lo-typ = typ-value(r-lo)
+    val r-hi-val-typ = typ-value(r-hi)
+  val vo-max = (v-in-max * r-lo-max / (r-lo-max + r-hi-val-min)) where :
+    val v-in-max = max-value(v-in)
+    val r-lo-max = max-value(r-lo)
+    val r-hi-val-min = min-value(r-hi)
+  
   min-typ-max(vo-min, vo-nom, vo-max)
```
ariel-hi, line 145: Ideally this could be written:
```
v-in * r-lo / (r-lo + r-hi)
```
However we're extracting max-value of `r-hi` when other values are lower, and vice versa.

ariel-hi:
Confirming behavior by runnning `quickstart-3`:
```
-> Tolerance 20.0%
    - Querying resistors for R1=390.0Ω R2=240.0Ω
      Checking solution for existing components with: tol1=MinMaxRange(min=-0.01, max=0.01), tol2=MinMaxRange(min=-0.005, max=0.005)
      Solved: mpn1=CRGCQ0402F390R, mpn2=ERA-2AED241X, v-out=(min=1.175V, nom=1.25V, max=1.325V), current=0.00520833333333333A
```
is shown before and after these changes.

modules/solvers/voltage-divider.stanza
```diff
@@ -25,23 +25,27 @@ public defstruct VoltageDividerSolution :
   vo: Toleranced
 
 public defn voltage-divider-solver (v-in: Toleranced,
-                             v-out: Toleranced,
-                             current: Double) -> VoltageDividerSolution|False:
+                                    v-out: Toleranced,
+                                    current: Double) -> VoltageDividerSolution|False:
```
bhusang, line 29: Should we also accept current as a Toleranced as well? `current: Toleranced|Double,`

There are situations where I want to spec a min and a max current through my voltage divider.

modules/solvers/voltage-divider.stanza
```diff
@@ -25,23 +25,27 @@ public defstruct VoltageDividerSolution :
...
   ; Look for highest acceptable tolerance in EIA standard values: [20% 10% 5% 2% 1% 0.5% 0.25% 0.1%]
   val current-tolerance = 0.1
-  val goal-r-hi = (typ-value(v-in) - typ-value(v-out)) / current
-  val goal-r-lo = typ-value(v-out) / current
+  val goal-r-hi = (v-in - v-out) / current
+  val goal-r-lo = v-out / current
```
bhusang, line 33: I assume that the arithmetic operators now handle the typ-value() extraction?

modules/solvers/voltage-divider.stanza
```diff
@@ -25,23 +25,27 @@ public defstruct VoltageDividerSolution :
...
   label<VoltageDividerSolution|False> return :
-    for tol in std-tols do :
+    for std-tol in std-tols do :
       ; std-tols are in % instead of being unitless
-      val tolerance = tol / 100.
+      val tolerance = std-tol / 100.
+      val lo-tol = tol(0.0, tolerance)
+      val hi-tol = tol(0.0, tolerance)
```
bhusang, line 39: This formulation allows negative tolerances which should be a no-no.

modules/solvers/voltage-divider.stanza
ariel-hi: Sure I'll support a Toleranced current. Perhaps `current` should always be Toleranced?

bhusang: Sure -- that's a bigger change but that is a better long term formulation.

modules/solvers/voltage-divider.stanza
ariel-hi: Before:
`goal-r-xx` variables were Doubles. The operation was divide(Double, Double).

PR:
`goal-r-xx` variables are Toleranced. The operation is divide(Toleranced, Double).

`typ-value(v-out) / current` is now a subset of the operations that happen when calculating the `goal-r-xx`.

modules/solvers/voltage-divider.stanza
ariel-hi: This commit should prohibit negative tolerances.
`std-tols` range up to 20.0.
`tolerance` will be at most `20.0 / 100.0 = 0.2`.
`tol(1.0, tolerance)` has min-value of at least 0.8

modules/solvers/voltage-divider.stanza
ariel-hi: I reverted this change. This PR is not intended to change the solution space, and changing the goals from Double to Toleranced was affecting that.

bhusang (approved):

modules/solvers/voltage-divider.stanza
m-hilgendorf: Is the current value a rating?

m-hilgendorf (approved):

PR #286, opened 2022-09-22
Update package for JITXValue

ariel-hi:
Depends on [#1506](https://github.com/JITx-Inc/jitx-client/pull/1506) in jitx-client.
JITXValue moved from `jitx/client` to `jitx/value`.


PhilippeFerreiraDeSousa (approved):
Approved even though the CI won't pass yet

PR #288, opened 2022-09-28
Added corner support for quad landpatterns

m-hilgendorf:


bhusang (approved):
With a comprehensive set of examples in the docs, this should be good to go.

PR #289, opened 2022-10-06
Break IO banks into separate symbol units in stm-to-jitx.

m-hilgendorf:


utils/stm.stanza
```diff
@@ -38,6 +38,23 @@ public pcb-struct ocdb/utils/stm/STMPinPropertiesRow :
   generic-props?: True|False
   power-props?: True|False
 
+public defn bank? (r:STMPinPropertiesRow) -> False|Ref :
+  val known-banks = [ "PA", 
+                      "PB", 
+                      "PC", 
+                      "PD" ]
```
bhusang, line 45: High pin count STM32 devices will be banks up to `H`.

m-hilgendorf: Fixed!

bhusang (approved):

PR #291, opened 2022-10-07
Updates to fix checks

m-hilgendorf:
- run-design has a final eval-when statement to force i/o checks after propagation of net-voltages
- changed description in checks to contain refs
- removed connectivity check for pins and added single-pin-net check
- removed conditional check generation, moved to info check in passive-checks/utils

### DIFF
--- a/designs/checks/capacitors.stanza
+++ b/designs/checks/capacitors.stanza
@@ -70,6 +70,121 @@ pcb-component ceramic-cap-info :
   symbol =
     sym(self.p[1] => sym.p[1], self.p[2] => sym.p[2])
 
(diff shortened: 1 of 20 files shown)


PhilippeFerreiraDeSousa:
Is there a CI test associated to the test design (/ example design) you are adding?

PhilippeFerreiraDeSousa (approved):
We should add a test on the `CheckSummary` returned by `run-checks`

```
public defstruct CheckSummary :
  filename:String
  num-passed:Int
  num-failed:Int
  num-incomplete:Int
  num-errors:Int
```

This would be in another PR

PR #295, opened 2022-10-18
inserted spaces after commas through a lot of files (per ocdb coding …

bhusang:
…standards)


m-hilgendorf (approved):
Assuming this passes CI, LGTM

bhusang:
It did not pass CI due to unknown types in the block:
```
  val r-lo = property(output-voltage-divider.r-lo.resistance)
  val true-adj-offset = adj-current * r-lo
```
Fixed.

PR #297, opened 2022-10-19
Update relative voltage propagation and unit tests

m-hilgendorf:
- Make sure all pins on a net have the same net voltage
- Add RV harness to unit tests


PhilippeFerreiraDeSousa (approved):
Interesting.
It would be nice to check that the eval-when condition ever becomes true

PR #305, opened 2022-11-18
Add two fixes to the voltage propagation utility

CuppoJava:
Add two fixes to the voltage propagation utility.
- Reworked run final passes to run in a more robust order. 
- Voltage propagation now checks for multiple conflicting voltages on the same net.


PhilippeFerreiraDeSousa (approved):
Thx for all the comments.
One hot take:
```
    val assigned-pins = assign-pins(design)
    val annotated-voltages = propagate-voltages(assigned-pins, voltage-prop-report)
    val new-design = run-evals?(annotated-voltages)
```
One may think putting each returned value in a named variable makes the code clearer, however it also opens the possibility for a typo by having pass 3 use output of pass1 instead of pass2. This is why I actually like at the end of the day
```
xs $> f
   $> g{_, arg2}
   $> h
```

SImilarly in match statements I use variable shadowing (usually bad practice) so that I do not mistakenly use the uncasted original variable
```
match (s) : 
  (s: T) : ...
```

Do you think this is a hot take?

PhilippeFerreiraDeSousa:
Note than an earlier version of this code was actually having this typo of pass_n using the output of an earlier pass by passing intermediate passes.

PR #310, opened 2022-11-29
check improved by only finding power-supply-pin pins

bhusang:
Also fixed the tests for io checks


d-haldane:
Can you target this at the `miro` branch, which has the rest of the updates to the checks?

bhusang:
Ok

PR #323, opened 2022-12-10
fixed locator pin holes to have no copper or soldermask (we have a st…

bhusang:
…upid way of doing the latter)


ptuckey (approved):
Thanks!!

PR #327, opened 2022-12-20
Expose Code IR

PhilippeFerreiraDeSousa:
Make IR public to be used by the post-processor instead of using raw JSON


ptuckey (approved):
LGTM!

PR #328, opened 2023-01-13
Add reference-prefix, text font, make standalone footprint

PhilippeFerreiraDeSousa:
How to use new Footprint constructor to visualize a standalone footprint:
```
view $ to-jitx $ Footprint(parse(slurp("mongo-component-database/parts-db/esir_footprints/<lp-uuid.json>")) as JObject)
```
json has `pads` and `landpattern` keys


ptuckey (approved):
Right on, thanks for adding this!

PR #329, opened 2023-01-13
fixed order of <hole, pad> for the 2-pin axial footprint

bhusang:


PhilippeFerreiraDeSousa:
I am on an branch  that branched out a few months ago and my ocdb has line 287
```
public defn pth-pad (pad-shape:Shape, hole-shape:Shape):
  pth-pad(pad-shape, false, false, hole-shape)
```

Your code has the arguments swapped.
```
public defn pth-pad (hole-shape:Shape, pad-shape:Shape) :
  pth-pad(hole-shape, pad-shape, false, false)
```

This is a very nasty breaking change for existing designs calling `pth-pad`, what's going on?

PhilippeFerreiraDeSousa:
I branched out from dev on Nov 22 on 2353078cef64328df809d0fa2278607fb655bd24 and the branch is dev_parts-db

bhusang:
Carl found that our argument order for through holes was inconsistent: sometimes we used (hole,pad) and other places used (pad,hole). In December, I made all of the pth calls consistent. Commit id: a280c57db19a21aba1edcc7838a21b9d0acfef9e

PhilippeFerreiraDeSousa (approved):

PR #330, opened 2023-01-14
Apply paste mask expansion

PhilippeFerreiraDeSousa:
Draft because not tested, python part requires more time


ptuckey:
Interesting!

Am I correct in thinking this is something we don't generally support in JITX atm? If so, maybe it's a general ticket to Triage?

PR #333, opened 2023-02-02
JITX-3925: Add PropertyCode

ptuckey:
This is the JITX parser code that pairs with the parts-db PR:
- [JITX-3925: Add PropertyCode](https://github.com/JITx-Inc/parts-db/pull/10)

(See there for test plan.)

### DIFF
--- a/utils/parts.stanza
+++ b/utils/parts.stanza
@@ -17,6 +17,8 @@ defpackage ocdb/utils/parts :
 ;============ Public API ==============
 ;======================================
 
+; We'll need an ESIR "Part <: Component" in order to support parametric queries.
+; Short-term, we're going directly into Instantiables.
 public defstruct PartCode :
   component: ComponentCode
 with :
@@ -41,7 +43,13 @@ public defstruct ComponentCode :
   symbols: Tuple<SymbolCode>
   datasheet: String
   image: String
+  trust: String
+  mounting: String
+  case: String|UNKNOWN
+  dimensions: DimensionsCode
+  sourcing: SourcingCode
   metadata: Tuple<MetadataCode>
+  properties: Tuple<ComponentPropertyCode>
   bundles:Tuple<BundleCode>
   supports:Tuple<SupportCode>
 with :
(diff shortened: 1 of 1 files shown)


erwin-lau (approved):
Should all the ???Code be subtype of ComponentCode?

ptuckey:
> Should all the ???Code be subtype of ComponentCode?

The `<Name>Code` convention is something we adopted during the prototype.

Generally, each `<Name>Code` represents a serializable version of an ESIR equivalent.

It's possible to instantiate a `<Name>Code` object, which could be useful for parametric queries (filtering before actually creating an `Instantiable`).

For now then, deserialization happens as a 2-step process:
1. Deserialize JSON as `<Name>Code` objects
2. Bring them into JITX via `to-jitx`, which returns an `Instantiable`

utils/parts.stanza
```diff
@@ -41,7 +43,13 @@ public defstruct ComponentCode :
   symbols: Tuple<SymbolCode>
   datasheet: String
   image: String
+  trust: String
+  mounting: String
+  case: String|UNKNOWN
+  dimensions: DimensionsCode
+  sourcing: SourcingCode
   metadata: Tuple<MetadataCode>
+  properties: Tuple<ComponentPropertyCode>
```
PhilippeFerreiraDeSousa, line 52: Why are you duplicating the queryable fields with the parametric API into the esir data Code?

PhilippeFerreiraDeSousa:
I am a bit confused by the mix of
- queryable data with the parametric query API
- esir data to reconstitute the part.

Currently you have
```
defstruct Part : 
  queryable-field1
  queryable-field2
  ...
code: Code
```
Code is the data to reconstitute the esir, it is not meant to be queried upon.

If you want to instantiate the part or do a `write-part-to-stanza-file(code(part))` you would take the code and make the part or export to code whatever.

Now at the root in the queryable field you can compute useful aggregates like `num-pins`, `num-gpio-pins`, `bundle-names` that a user would like to query. It is not very useful to have the nested json with raw data about esir things mixed with queryable fields.

Finally here is how you would be restraining a parametric query to only return parts with esir data:
```
chip-resistor(["resistance" => 3.0, "_exist" => "code"])
```

PhilippeFerreiraDeSousa (approved):

ptuckey:
> I am a bit confused by the mix of
> 
> * queryable data with the parametric query API
> * esir data to reconstitute the part.

Thanks for the callout, Philippe!

We chatted about this over Slack.

I agree that it's confusing to be writing the same piece of data in more than one place.

For now, this PR supports parsing in both places (which hands over control to the data pipeline).

PR #334, opened 2023-02-03
JITX-3924: Add part categories

ptuckey:
Add parsing support for JITX-specific categories.

### DIFF
--- a/utils/parts.stanza
+++ b/utils/parts.stanza
@@ -391,7 +391,7 @@ defn string-or-unknown (json: JSON) -> String|UNKNOWN :
     (s: JNull) :
       UNKNOWN
 
-public defenum Category :
+defenum Category :
   capacitor
   connector
   diode
@@ -399,7 +399,7 @@ public defenum Category :
   inductor
   microcontroller
   oscillator
-  power_supply_chip
+  power-supply-chip
   resistor
   transistor
 
@@ -412,7 +412,7 @@ defn category-or-unknown (category-str: String) -> Category|UNKNOWN :
     "inductor": inductor
     "microcontroller": microcontroller
     "oscillator": oscillator
-    "power_supply_chip": power_supply_chip
+    "power_supply_chip": power-supply-chip
     "resistor": resistor
     "transistor": transistor
     else: UNKNOWN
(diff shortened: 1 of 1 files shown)


utils/parts.stanza
```diff
@@ -389,6 +391,32 @@ defn string-or-unknown (json: JSON) -> String|UNKNOWN :
...
+    "connector": connector
+    "diode": diode
+    "fuse": fuse
+    "inductor": inductor
+    "microcontroller": microcontroller
+    "oscillator": oscillator
+    "power_supply_chip": power_supply_chip
+    "resistor": resistor
+    "transistor": transistor
+    else: UNKNOWN
```
PhilippeFerreiraDeSousa, line 418: Hey, here is a compact way to write it, it is less readable but has lower maintenance / less error prone:
```
import lang-utils

val category-values = map(Category, 0 to Category-length)
val string-to-category-value = to-hashtable(to-string, category-values)

defn category-or-unknown (category-str: String) -> Category|UNKNOWN :
  get?(string-to-category-value, category-str, UNKNOWN)
```

For OCDB, preferring readability like you did makes sense, so not a change request

PhilippeFerreiraDeSousa (approved):

utils/parts.stanza
```diff
@@ -389,6 +391,32 @@ defn string-or-unknown (json: JSON) -> String|UNKNOWN :
...
 
+public defenum Category :
+  capacitor
+  connector
+  diode
+  fuse
+  inductor
+  microcontroller
+  oscillator
+  power_supply_chip
```
PhilippeFerreiraDeSousa, line 402: Do we want this with `-` instead of `_`?

utils/parts.stanza
ptuckey: Oh awesome thanks! I was wondering if there was a better way...

I'll leave as is for the reasons you stated.

utils/parts.stanza
ptuckey: Good call!

The fields within the schema itself all use underscores (to be Python-friendly) - but, here it makes sense to use Stanza idioms.

PR #335, opened 2023-02-03
JITX-3924: Add vendor part numbers

ptuckey:
This passes through a map of vendor-specific part numbers,
which can be used for BOM generation.

It also brings in support for `KeyValue` properties, which is how
so it can be displayed in the design explorer.


erwin-lau (approved):
LGTM

PhilippeFerreiraDeSousa (approved):

PR #341, opened 2023-02-15
JITX-4098: Support filtering by vendor part number availability

ptuckey:
## 1. Remove `lcsc` as a hardcoded vendor key
- Use a HashTable where any String can be a vendor key
- Allow adding in further vendors without having to update schema/client

## 2. Include an `exists` boolean
Allow parametric queries to filter to parts that include this key:
```
database-part(["vendor_part_numbers.lcsc.exists" => true])
```

We already support `_sellers` behavior for filtering by offers - so, it may turn out we don't
need this particular filter in the long run - but, it gives us the option, as we test/learn.

### DIFF
--- a/utils/parts.stanza
+++ b/utils/parts.stanza
@@ -22,7 +22,7 @@ public defstruct ComponentCode :
   description: String
   manufacturer: String
   mpn: String
-  vendor-part-numbers: VendorPartNumbersCode
+  vendor-part-numbers: HashTable<String, VendorPartNumberCode>
   category: Category|UNKNOWN
   emodel: EModel|False
   pin-properties: False|PinPropertiesCode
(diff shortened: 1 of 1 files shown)


utils/parts.stanza
```diff
@@ -141,8 +141,9 @@ public-when(TESTING) defstruct NoConnectCode :
 with :
   printer => true
 
-public-when(TESTING) defstruct VendorPartNumbersCode :
-  lcsc: String|UNKNOWN
+public-when(TESTING) defstruct VendorPartNumberCode :
+  exists: True
```
ariel-hi, line 145: If the `exists` field is always True maybe it is not needed?

ariel-hi: Reading your PR description I think I understand better.
Perhaps use a `True|False` type even if it is only populated with True for now?

ptuckey: :+1:

ariel-hi (approved):
Great!

erwin-lau (approved):
Yep, that should work!

PR #347, opened 2023-02-17
JITX-4098: vendor_part_numbers - update to use "_exists" API

ptuckey:
Simplify the PartsDB schema by removing unnecessary `exists/value` fields.

Thanks @PhilippeFerreiraDeSousa for pointing out that the API already supports `_exists`!

```
# Before
database-part(["category" => "resistor", "vendor_part_numbers.lcsc.exists" => true])

# After
database-part(["category" => "resistor", "_exist" => ["vendor_part_numbers.lcsc"]])
```

### DIFF
--- a/utils/parts.stanza
+++ b/utils/parts.stanza
@@ -40,7 +40,7 @@ public defstruct ComponentCode :
   description: String
   manufacturer: String
   mpn: String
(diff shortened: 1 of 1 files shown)


erwin-lau (approved):
Congrats!

PR #350, opened 2023-02-23
JITX-4095: Fix tolerance parsing for parts-db

ptuckey:
Was looking for the wrong name - ran into this while testing a parts-db change.


erwin-lau (approved):
:o

PR #351, opened 2023-02-28
JITX-4095: Wrap component with queryable parts

ptuckey:
## Summary
This goes along with an accompanying [parts-db PR](https://github.com/JITx-Inc/parts-db/pull/36).

The purpose is to allow proper parametric queries, such as:
```
database-part(["category" => "resistor", "resistance" => 7500])
```

### Component Wrapping
In order to achieve this, the top-level parts are category-specific, queryable documents,
containing a `component` (the approach @PhilippeFerreiraDeSousa originally took).

### Migration Backwards-Compability
Additionally, this includes support to transform a parts-db result into one of our existing queryable structs:
- Resistor
- Capacitor
- Inductor

This is for backwards-compatibility, as we migrate the data to the parts-db.

Note that existing queries like `chip-resistor()` are currently unaffected, and still load from the old DB.

We can switch those over as a follow-up.

### DIFF
--- a/utils/db-parts.stanza
+++ b/utils/db-parts.stanza
@@ -17,6 +17,7 @@ defpackage ocdb/utils/db-parts :
   import ocdb/utils/defaults
   import ocdb/utils/landpatterns
   import ocdb/utils/generator-utils
+  import ocdb/utils/parts
   import ocdb/utils/property-structs
   import ocdb/utils/symbols
   import ocdb/utils/design-vars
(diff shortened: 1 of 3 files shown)


utils/db-parts.stanza
```diff
@@ -296,18 +319,33 @@ public defn Capacitor (raw-json: JObject) -> Capacitor :
...
+  val mounting = if get?(json, "mounting") is String:
+    json["mounting"] as String
+  else:
+    ""
+
+  val vendor-part-numbers = if get?(json, "vendor_part_numbers") is JObject:
+    VendorPartNumbers(json["vendor_part_numbers"] as JObject)
+  else:
+    []
+
```
erwin-lau, line 336: I would slightly prefer it written this way:
```
val component = match(get?(json, "component")) :
  (j:JObject) :
    ComponentCode(j)
  (_) : false

val mounting = match(get?(json, "mounting")) :
  (s:String) : s
  (_) : ""

val vendor-part-numbers = match(get?(json, "vendor_part_numbers")) :
  (j:JObject) : VendorPartNumbers(j)
  (_) : []

erwin-lau (approved):
LGTM! (I am not a bot)

utils/db-parts.stanza
ptuckey: Thanks - much nicer!

PR #352, opened 2023-03-01
JITX-4209: Update "is-resistor()" to check "category"

ptuckey:
The parts-db is includes parts with "resistance", that shouldn't be considered resistors,
for validation/checking purposes.

For example, a potentiometer can have a `resistance`, but can have more than 2 pins.

This changes `is-resistor()` to check by `category` instead (and similar for capacitor/inductor).

## Test Plan
Used one of @CaydenPierce's projects which was failing previously:
```
FATAL ERROR: Resistor, capacitor or inductor amp.vol has 5 pins instead of 2.
  in ocdb/utils/generator-utils/calculate-operating-points
    at open-components-database/utils/generator-utils.stanza:172.4
  in ocdb/utils/generator-utils/calculate-operating-points
    at open-components-database/utils/generator-utils.stanza:175.39
  in ocdb/utils/generator-utils/run-final-passes
    at open-components-database/utils/generator-utils.stanza:291.22
  in helpers/run-check-on-design
    at helpers.stanza:20.20
```

Ran with this branch and confirmed it works.

### DIFF
--- a/utils/db-parts.stanza
+++ b/utils/db-parts.stanza
@@ -176,7 +176,10 @@ defmethod to-jitx (resistor: Resistor) -> Instantiable :
     val rated-temperature = rated-temperature(resistor)
     val tolerance = tolerance(resistor)
     val resistance = resistance(resistor)
+    ; Backwards compatibility
     property(self.resistor) = true
+    ; This is how parts-db does it
+    property(self.category) = "resistor"
     property(self.trust) = trust(resistor)
     property(self.x) = x(resistor)
     property(self.y) = y(resistor)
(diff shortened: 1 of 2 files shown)


erwin-lau (approved):
But what if a component could fall into many categories?

ptuckey:
> But what if a component could fall into many categories?

Good question!

Was talking with @bhusang yesterday about potentially adding sub-categories.

For now, we're defining `resistor` as just 2-pin resistors, which should be backwards-compatible with calls like `chip-resistor(...)`.

PR #355, opened 2023-03-03
JITX-4238: Hack to insert ">REF" and ">VALUE" for all symbols

ptuckey:
The parts-db component symbols aren't inserting ">REF" and ">VALUE" Text symbols.

These are required for:
- Rendering in JITX view
- Exporting to KiCad (so it shows up as "Value" in "Symbol Properties"

In the short term, we're hacking this into OCDB deserialization, but the plan is to
fix it properly by inserting these during parts-db population.


erwin-lau (approved):
Yeah!

emspin (approved):

PR #359, opened 2023-03-14
Usb c cable tester - add usb C full pinout bundle, improve USB pinout for USB-C cable tester

CaydenPierce:


utils/bundles.stanza
```diff
@@ -185,6 +185,27 @@ public pcb-bundle usb-c :
...
+  pin GND
+  pin shield
+  pin SSTXP1
+  pin SSTXN1
+  pin SSRXP1
+  pin SSRXN1
+  pin SSRXP2
+  pin SSRXN2
+  pin SSTXP2
+  pin SSTXN2
```
bhusang, line 208: There is no reason to do the following:
```
port CC : pin [[1 2]]
port DP : pin [[1 2]]
```
and so on...

bhusang: That way you could refer to CC as CC[1] and CC[2]. Not sure of the utility but it does reduce the number of lines. And it allows the user to access 2 pins with a loop.

bhusang (approved):
Think about indexing the -1 and -2 signals. Other than that -- looks good.

utils/bundles.stanza
CaydenPierce: I think I will hold off on doing this because:
1. This will break the net name extraction, unfortunately (doesn't work for array pins).
2. This will change any loop over the pins to require multiple levels of looping. It actually appears to complicate the usage of the bundle (at least in my usage).

PR #360, opened 2023-03-14
parametric resistors and capacitors (first pass resolution)

bhusang:


bhusang:
This PR also addresses part 1 of JITX-4121 to use EIA values that are more commonly available.

components/nexperia/PMBT3904.stanza
```diff
@@ -28,62 +28,62 @@ public pcb-component component :
   symbol = sym(b => sym.b, c => sym.c, e => sym.e)
   property(self.rated-temperature) = min-max(-65.0, 150.0)
 
-  spice :
```
PhilippeFerreiraDeSousa, line 31: Are you sure you want to merge those commented out blocks onto `dev` ?

utils/generic-components.stanza
```diff
@@ -163,6 +163,43 @@ defn smd-query-properties () -> Tuple<KeyValue<String, ?>> :
 public defn remove-duplicate-keys (hs:Seqable<Seqable<KeyValue<String, ?>>>) -> Tuple<KeyValue<String, ?>> :
   to-tuple $ to-hashtable<String, ?>(cat-all(hs))
 
+; ========================================================
+; ========== Handle min/max value searching ==============
+; ========================================================
+
+defn kvs<?K,?V> (xs:Seqable<KeyValue<?K,?V>>) :
+  for x in xs seq : [key(x), value(x)]
```
PhilippeFerreiraDeSousa, line 171: You don't want to `import lang-utils` ?

utils/generic-components.stanza
```diff
@@ -163,6 +163,43 @@ defn smd-query-properties () -> Tuple<KeyValue<String, ?>> :
...
+                yield("max-resistance" => max-value(myv))
+              "capacitance" :
+                yield("min-capacitance" => min-value(myv))
+                yield("max-capacitance" => max-value(myv))
+              "inductance" :
+                yield("min-inductance" => min-value(myv))
+                yield("max-inductance" => max-value(myv))
+          (_) :
+            yield(myk => myv)          
+  replaced-params
```
PhilippeFerreiraDeSousa, line 193:
```
defn replace-params (params:Tuple<KeyValue>) -> Tuple<KeyValue> :
  ; need to check if resistance is passed a Toleranced() value and if
  ; so, transform it to min-resistance, max-resistance
  to-tuple $
    for [myk, myv] in kvs(params) seq-cat :
      match(myv:Toleranced) :
        [append("min-", myk) => min-value(myv),
         append("max-", myk) => max-value(myv)]
      else :
        [myk => myv]
```

utils/generic-components.stanza
```diff
@@ -163,6 +163,43 @@ defn smd-query-properties () -> Tuple<KeyValue<String, ?>> :
...
+                yield("min-capacitance" => min-value(myv))
+                yield("max-capacitance" => max-value(myv))
+              "inductance" :
+                yield("min-inductance" => min-value(myv))
+                yield("max-inductance" => max-value(myv))
+          (_) :
+            yield(myk => myv)          
+  replaced-params  
+
+defn lookup-val (params:Tuple<KeyValue>, key:String) -> Double|Toleranced|False :
```
PhilippeFerreiraDeSousa, line 195: `value $ find!({/key(_) == key}, params)` but usually you need to call such function several time so just make an hashtable out of params

utils/generic-components.stanza
bhusang: Ok, I can do that. Thanks!

utils/generic-components.stanza
```diff
@@ -173,19 +210,28 @@ public defn ceramic-cap (user-params:Tuple<KeyValue>) -> Instantiable :
   val params = remove-duplicate-keys $ [["type" => "ceramic" "min-rated-voltage" => MIN-CAP-VOLTAGE]
                                         smd-query-properties()
                                         user-params]
-  to-jitx $ Capacitor(params)
-
-public defn ceramic-cap (cap:Double) -> Instantiable :
+  val my-params = replace-params(params)
```
PhilippeFerreiraDeSousa, line 213: `replace-params` crashes if another key than `resistance` / `inductance` / `capacitance` has a value of type `Toleranced`. You cannot make this assumption here.

utils/generic-components.stanza
```diff
@@ -173,19 +210,28 @@ public defn ceramic-cap (user-params:Tuple<KeyValue>) -> Instantiable :
...
+  try :
+    to-jitx $ Capacitor(my-params)
+  catch (e) :
+    val my-cap = lookup-val(params, "capacitance")
+    println("Warning: generic capacitor substitution for value (%_)" % [my-cap])
+    match(my-cap) :
+      (d:Double) : dummy-capacitor(d)
+      (t:Toleranced) : dummy-capacitor(t)
+      (_) : fatal("Did not find acceptable type of capacitance value (%_)" % [my-cap])
```
PhilippeFerreiraDeSousa, line 222: Do you want to specifically catch `jitx/errors/NoComponentMeetingRequirements` exceptions?
Errors from the database can be of:
- NoComponentMeetingRequirements
- InvalidPartDatabaseQueryAttributeType
- InvalidPartDatabaseQuerySpecialKey
- InvalidPartDatabaseQueryProperty
- HTTPError

utils/generic-components.stanza
bhusang: There could be other parameters that may be toleranced that I don't want to replace, just the keys of resistance, capacitance, and inductance. I will modify your cool suggestion, though.

utils/generic-components.stanza
```diff
@@ -208,19 +254,28 @@ public defn tantalum-cap (user-params:Tuple<KeyValue>) -> Instantiable :
   val params = remove-duplicate-keys $ [["anode" => "tantalum" "min-rated-voltage" => MIN-CAP-VOLTAGE]
                                         smd-query-properties()
                                         user-params]
-  to-jitx $ Capacitor(params)
-
-public defn tantalum-cap (cap:Double) -> Instantiable :
+  val my-params = replace-params(params)
```
PhilippeFerreiraDeSousa, line 257: `replace-params` assumptions not met

utils/parametric-components.stanza
```diff
@@ -0,0 +1,537 @@
...
+  ["mounting" => "smd"
+   "case" => get-valid-pkg-list()
+   "minimum_quantity" => 1
+   "min-stock" => 5 * DESIGN-QUANTITY]
+
+defn app-query-properties (app:CapacitorApplication) -> Tuple<KeyValue<String, ?>> :
+  switch(app) :
+    CapBypass : ["tolerance" => 0.10]
+
+defn kvs<?K,?V> (xs:Seqable<KeyValue<?K,?V>>) :
```
PhilippeFerreiraDeSousa, line 103: Again?

utils/parametric-components.stanza
```diff
@@ -0,0 +1,537 @@
...
+  else :
+    property(self.orig-inductance) = ind
+  ; spice :
+  ;   "[L] <p[1]> <p[2]> <ind>"
+
+defn smd-query-properties () -> Tuple<KeyValue<String, ?>> :
+  ["mounting" => "smd"
+   "case" => get-valid-pkg-list()
+   "minimum_quantity" => 1
+   "min-stock" => 5 * DESIGN-QUANTITY]
```
PhilippeFerreiraDeSousa, line 97: You may want to deduplicate with the slightly different function in generic-components or not use the same function name
```
defn smd-query-properties () -> Tuple<KeyValue<String, ?>> :
  ["mounting" => "smd"
   "case" => get-valid-pkg-list()
   "min-stock" => 5 * DESIGN-QUANTITY]
```

utils/parametric-components.stanza
```diff
@@ -0,0 +1,537 @@
...
+
+  ; Returns a value within the range of v in the tolerance range specified by tol
+public defn between-std-val (v:Toleranced, tol:Double) -> Double :
+  val expon-min = floor(log10(min-value(v)))
+  val expon-max = floor(log10(max-value(v)))    
+  var match-exp = 0.0
+  var match-val = -1.0
+  var match-diff = 1.0e6
+  ; println("Need to find exp range (%_ %_) for range (%_)" % [expon-min, expon-max, v])
+  for exp in to-int(expon-min) to to-int(expon-max + 1.0) do :
```
PhilippeFerreiraDeSousa, line 355: You can use `through N` instead of `to (N + 1)

components/nexperia/PMBT3904.stanza
bhusang: For now, I think so -- it does require a change to jitx-client to support it but I wanted to keep the code ready to go.

utils/generic-components.stanza
bhusang: I guard against that in the updated replace-params() function.

utils/generic-components.stanza
bhusang: I also want to catch the other errors you listed because we don't want the design to crash. We print a warning when we use a dummy-* component so that the user can continue compiling the design even if no internet connectivity. With the warning, the user can go back and modify their code if they want real components (fix query, wait for internet connectivity).

utils/parametric-components.stanza
bhusang: Good suggestion. I made the one in generic-components public and deleted this defn.

utils/parametric-components.stanza
bhusang: Be gone!

utils/parametric-components.stanza
bhusang: 😃 Good suggestion

utils/generic-components.stanza
bhusang: Fixed.

utils/parametric-components.stanza
```diff
@@ -0,0 +1,537 @@
...
+
+; ========================================================
+; ========== parametric part helper functions ============
+; ========================================================
+defn c-r (v:Double e:Int) :
+  v * pow(10.0, to-double(e))
+
+
+  ; Returns a value within the range of v in the tolerance range specified by tol
+public defn between-std-val (v:Toleranced, tol:Double) -> Double :
```
PhilippeFerreiraDeSousa, line 348: Why is `tol` not the `Toleranced` and `v` the Double value but the other way?

utils/parametric-components.stanza
```diff
@@ -0,0 +1,537 @@
...
+  val expon-min = floor(log10(min-value(v)))
+  val expon-max = floor(log10(max-value(v)))    
+  var match-exp = 0.0
+  var match-val = -1.0
```
PhilippeFerreiraDeSousa, line 352: You will get those made up values out if there are no pairs in the range, do you want to use closest-std-val as fallback?

utils/parametric-components.stanza
```diff
@@ -0,0 +1,537 @@
...
+    for std-val in union-std-vals(tol) as List<Double> do :
+      ; println("in-range?(%_, %_)" % [v, c-r(std-val, exp)])
+      if in-range?(v, c-r(std-val, exp)) :
+        val diff = abs(c-r(std-val, exp) - center-value(v))
+        if diff < match-diff :
+          ; std-val component (at the appropriate exp range) is closest to center()
+          match-diff = diff
+          match-val = std-val
+          match-exp = to-double(exp)
+  match-val * pow(10.0, match-exp)
```
PhilippeFerreiraDeSousa, line 365: Here is a proposal
```
public defn between-std-val (v:Toleranced, tol:Double) -> Double :
  val DEFAULT-VALUE = [-1.0, 0.0]
  val expon-min = floor(log10(min-value(v)))
  val expon-max = floor(log10(max-value(v)))

  defn valid? (args: [Double, Int]) -> True|False :
    in-range?(v, make-value(args))

  defn distance-to-center (args: [Double, Int]) -> Double :
    abs(make-value(args) - center-value(v))

  defn make-value (args: [Double, Int]) -> Double :
    val [std-val, exp] = args
    c-r(std-val, exp)

  defn c-r (v:Double e:Int) -> Double :
    v * pow(10.0, to-double(e))

  val exp-range = to-int(expon-min) through to-int(expon-max)

  val valid-pairs = to-tuple $ filter(valid?, zip(union-std-vals(tol), exp-range))

  if empty?(valid-pairs) :
    val closest-std-value = closest-std-val(typ(v), tol)
    println("No standard value matching request v=%_, tol=%_. Using closest std value." % [v, tol, closest-std-value])
    closest-std-value
  else :
    make-value $ minimum(distance-to-center, valid-pairs)
```
`closest-std-val` takes 2 Double, it is possible I am swapping the arguments

utils/parametric-components.stanza
bhusang: Just a quirk of how tolerance for components is handled --> it's a double which signifies the the 20%, 10%, 5%, 2%, 1% etc available ranges for component families.

utils/parametric-components.stanza
```diff
@@ -0,0 +1,537 @@
...
+
+public defn find-std-val (v:Toleranced) -> Double:
+  var done = 0
+  var ans = 0.0
+  val my-val = 
+    label<Double> return :
+      for tol in std-tols do :
+        ans = between-std-val(v, tol)
+        ; println("Debug tol(%_) ans(%_)" % [tol, ans])
+        return(ans) when ans > 0.0
```
PhilippeFerreiraDeSousa, line 375: Oh, are you using `-1` as default in the function above as a hack to discard it here? You can just return `Maybe<Double>`. Then:
```
val my-val = value! $ first?(between-std-val{v, _}, std-told)
```

utils/parametric-components.stanza
bhusang: Kind of -- we want to find the std value that in the range of the mantissa but don't pay attention to the exponent.

utils/parametric-components.stanza
```diff
@@ -0,0 +1,537 @@
...
+    new-comps
+
+  within transform-module(m) :
+    inside pcb-module :
+      val knownR = HashTable<Double|Toleranced, Vector<JITXObject>>()
+      val knownC = HashTable<Double|Toleranced, Vector<JITXObject>>()
+      val knownL = HashTable<Double|Toleranced, Vector<JITXObject>>()
+      val paraR = HashTable<Double|Toleranced, Vector<JITXObject>>()
+      val paraC = HashTable<Double|Toleranced, Vector<JITXObject>>()
+      val paraL = HashTable<Double|Toleranced, Vector<JITXObject>>()
```
PhilippeFerreiraDeSousa, line 462: You can create them with `HashTable-init<Double|Toleranced, Vector<JITXObject>>(Vector<JITXObject>{})`.

This way `table[k]` will initialize the value to an empty vector first if needed.

utils/parametric-components.stanza
```diff
@@ -0,0 +1,537 @@
...
+      for i in list-all-instances() do :
+        if has-property?(i.parametric) :
+          ; println("Sorting part %_ with desired value %_" % [ref(i) property(i.resistance)])
+          if has-property?(i.resistor) :
+            match(get?(paraR, property(i.resistance))) : 
+              (v:Vector<JITXObject>) : add(v,i)
+              (_) : 
+                val t = Vector<JITXObject>()
+                add(t,i)
+                set(paraR, property(i.resistance), t)
```
PhilippeFerreiraDeSousa, line 475:
```
if has-property?(i.resistor) :
  add(paraR[property(i.resistance)], t)
```
after using `HashTable-init`

bhusang: Very cool.

utils/parametric-components.stanza
```diff
@@ -0,0 +1,537 @@
...
+              set(knownL, c, t)
+    
+      ; we have 2n Hashtables here 
+      ; paraR,paraC,.. are the parametric components
+      ; knownR,knownC,... are the known fixed components (non-parametric)
+      resolve-components(paraR, knownR)
+      resolve-components(paraC, knownC)
+      resolve-components(paraL, knownL)
+
+public defn list-all-instances () -> Vector<JITXObject> :
```
PhilippeFerreiraDeSousa, line 527: In short
```
public defn list-all-instances () -> Seqable<JITXObject> :
  let loop (m:JITXObject = self) :
    for i in single-instances(m) seq-cat :
      cat([i], loop(i))
```

PhilippeFerreiraDeSousa: Surprisingly you don't need the `inside pcb-module`.
I tried it with:
```
#use-added-syntax(jitx)
defpackage ocdb/designs/grid-resistors :
  import core

  import jitx
  import jitx/commands
  import ocdb/utils/defaults
  import ocdb/utils/generic-components

val board-shape = RoundedRectangle(10.0, 10.0, 0.5)

public defn list-all-instances () -> Seqable<JITXObject> :
  let loop (m:JITXObject = self) :
    for i in single-instances(m) seq-cat :
      cat([i], loop(i))

pcb-module main :
  inst i : database-part(["vendor_part_numbers.lcsc" => "C2290"])

  println $ to-tuple $ list-all-instances()

defn run-design (circuit: Instantiable) :
  make-default-board(circuit, 2, board-shape)

run-design(main)
```

```
Current working directory: "/Users/philippe/dev/jitx-client"
Loading JITX Shell...
[Instance(L294/D265(relative))]
```

bhusang: Nice

utils/parametric-components.stanza
bhusang: So I don't quite understand how to integrate your code.
```
      for i in list-all-instances() do :
        if has-property?(i.parametric) :
          ; println("Sorting part %_ with desired value %_" % [ref(i) property(i.resistance)])
          if has-property?(i.resistor) :
            add(paraR[property(i.resistance)], t)
```
where does the `t` come from? If the key (i.resistance) doesn't exist the old code created `t` as a vector which became the value for the hashtable.

utils/generic-components.stanza
PhilippeFerreiraDeSousa: Ok, however you are silently catching the error and printing a hardcoded different (potentially wrong) error message to the terminal.
You need to still print the error.
The typo checker has elaborate error printers listing the available attributes and special keys when the user makes a typo.
The typo checker is actually meant to stop execution and let the user fix the query.

utils/parametric-components.stanza
PhilippeFerreiraDeSousa: Sorry, `i` not `t`.
```
if has-property?(i.resistor) :
  add(paraR[property(i.resistance)], i)
```

PhilippeFerreiraDeSousa: Just assume there is already a vector in `paraR[key]`, for any `res`. So you can safely append to it.

PhilippeFerreiraDeSousa (approved):
Hey, approving so that you can change what you want and merge when you want.
Have a good weekend!

utils/parametric-components.stanza
bhusang: Cool.

utils/generic-components.stanza
bhusang: Ok, will now only catch HTTPError & NoComponentMeetingRequirements. The remainder will trigger the in-place error handlers.

utils/parametric-components.stanza
bhusang: Accepted your proposal. Looks good in my testing (limited).

PhilippeFerreiraDeSousa:
The CI is green

PR #368, opened 2023-03-27
replace set-design-directory with set-current-design

erwin-lau:
# Changes
* replace all `set-design-directory` command calls with `set-current-design`
* remove trailing spaces in some cases.

# Notes
* some designs use `altium` some `kicad`. They are left as is.


PhilippeFerreiraDeSousa (approved):
Nice

ptuckey (approved):
Gogogo!

PR #372, opened 2023-03-28
updated all instances of mentioned datasheets. Note that all componen…

bhusang:
…ts do not necessarily have a datasheet.

### DIFF
--- a/components/espressif/ESP32-PICO-D4.stanza
+++ b/components/espressif/ESP32-PICO-D4.stanza
@@ -14,9 +14,7 @@ defpackage ocdb/components/espressif/ESP32-PICO-D4 :
   import ocdb/utils/bundles
   import ocdb/utils/box-symbol
   import ocdb/utils/property-structs
-
-; datasheet reference: 
-
+  
 public pcb-component component :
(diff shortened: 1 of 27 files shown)


bhusang:
Nice to get the `datasheet` tab filled in when you do the following:
```
view-design-explorer $ component
```
after running the repl on the component file.

components/espressif/ESP32-PICO-D4.stanza
```diff
@@ -15,11 +15,12 @@ defpackage ocdb/components/espressif/ESP32-PICO-D4 :
   import ocdb/utils/box-symbol
   import ocdb/utils/property-structs
 
-; datasheet reference: https://www.espressif.com/sites/default/files/documentation/esp32-pico-d4_datasheet_en.pdf
+; datasheet reference:
```
d-haldane, line 18: Stray comment line hanging out here

d-haldane (approved):

components/espressif/ESP32-PICO-D4.stanza
bhusang: Oops.

PR #379, opened 2023-04-12
Update ocdb/main in prep for GA of jitx-client

bhusang:
Updating ocdb/main so that another PR on jitx-client will update OCDB to the latest version for customers.


ptuckey (approved):
Yep!!

PhilippeFerreiraDeSousa (changes requested):
Let the CI fast-forward `main` on public release

bhusang:
Does the CI fast forward from dev then?

PR #385, opened 2023-04-22
Quickstart updates

CaydenPierce:
Banana plug labels weren't working - swap to use `text` so they are now working properly. Must get in to align with quickstarts.


CaydenPierce:
Not meant for main - changed to dev

bhusang (approved):

PR #387, opened 2023-04-24
Reduce ocdb altium symbol size by 254 - adjustments

PhilippeFerreiraDeSousa:
See change in `altium-power-gnd-signal-sym`

# Before

image

# After

image

Here is the Altium reference
image


erwin-lau (approved):
Sorry, I missed this one.

PR #394, opened 2023-05-08
JITX-4967/npth-pad

erwin-lau:
# Background

* DRC Error because of NPTH Pad inside an SMD Pad

# Changes
* Modify the offending component with PTH Pads instead of using Cutouts

# Further Works
There may be more components that need this treatment as well as during import (CAD or parts-db)

### DIFF
--- a/components/johnson/142-0761-881.stanza
+++ b/components/johnson/142-0761-881.stanza
@@ -23,25 +23,23 @@ public pcb-component component :
   pin-properties :
     [pin:Ref     | pads:Int ...   | side:Dir]
     [sig | 1 | Left]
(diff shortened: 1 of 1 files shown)


ptuckey (approved):
Is there a way of checking for this within JITX itself, so the user doesn't have to get as far as DRC in external CAD?

erwin-lau:
> Is there a way of checking for this within JITX itself, so the user doesn't have to get as far as DRC in external CAD?

Unfortunately no. From a layman perspective, it is a valid geometry. Only when it goes to manufacturing and considering how to solder a component you realize this is not correct. In JITX, there is no check of this kind as far as I know.

PR #398, opened 2023-05-18
JITX-5031: Map "stock" -> "min-stock"

ptuckey:
If a user's parametric query includes "stock", interpret as "min-stock".

Other changes:
- Move remapping logic from jitx-client to OCDB, so it runs before deduplication.
- Change existing deduplication calls to a wrapping normalization function.

## Test Plan
```
import ocdb/utils/generic-comnponents

; Passes
chip-resistor(["resistance" => 13500.0 "stock" => 5])
chip-resistor(["resistance" => 13500.0 "min-stock" => 5])

; Fails, due to stock
chip-resistor(["resistance" => 13500.0 "stock" => 500000])
chip-resistor(["resistance" => 13500.0 "min-stock" => 500000])
```

### DIFF
--- a/utils/parametric-components.stanza
+++ b/utils/parametric-components.stanza
@@ -338,7 +338,7 @@ public defn normalize-params (hs:Seqable<Seqable<KeyValue<String, ?>>>) -> Tuple
 
 ; Substitute user params to match expectations:
 ;   - https://en.wikipedia.org/wiki/Principle_of_least_astonishment
-defn substitute-params (params: Seqable<KeyValue<String, ?>>) -> Seqable<KeyValue<String, ?>> :
+defn substitute-params (params: Seqable<KeyValue<String, ?>>) -> Tuple<KeyValue<String, ?>> :
   to-tuple $ for kv in params seq :
     switch {key(kv) == _} :
       "minimum_quantity" : "max-minimum_quantity" => value(kv)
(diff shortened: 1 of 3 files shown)


utils/parametric-components.stanza
```diff
@@ -332,6 +332,23 @@ public defn parametric-resistor (res:Double|Toleranced, tol:Double) -> Instantia
...
+
+; Substitute user params to match expectations:
+;   - https://en.wikipedia.org/wiki/Principle_of_least_astonishment
+defn substitute-params (params: Seqable<KeyValue<String, ?>>) -> Seqable<KeyValue<String, ?>> :
+  to-tuple $ for kv in params seq :
+    switch {key(kv) == _} :
+      "minimum_quantity" : "max-minimum_quantity" => value(kv)
+      "stock" : "min-stock" => value(kv)
+      else :
+        kv
```
erwin-lau, line 347: Maybe it is better to return things  as `Tuple` and use `Seqable` as argument

erwin-lau (approved):
That makes sense

utils/parametric-components.stanza
ptuckey: Sounds good!

ptuckey: Oh, it's already calling `to-tuple` above - the signature was just typed to `Seqable` - I'll update that to match `Tuple`.

Thanks!

PR #403, opened 2023-06-12
JITX-5178: Add sellers data as property for component card

ptuckey:
Ensure that sellers data displays in component cards.

This allows users to see offers, where the data comes from either Parts DB, or Octopart.

## Test Plan
```
import ocdb/utils/generic-components

; Including "_stock" causes fresh data to be fetched from Octopart
view-design-explorer $ database-part(["mpn" => "RP2040", "_stock" => 0])
```

### DIFF
--- a/utils/parts.stanza
+++ b/utils/parts.stanza
@@ -51,7 +51,7 @@ public defstruct Part <: Component :
   ; Power dissipation limit as rated by manufacturer. One value, or PWL function (W | [degC, W])
   rated-power: Double|False
 
-  ; Optionally appended by jit-backend (does not come from parts-db)
+  ; These may be populated in Parts DB, or updated values may be inserted by the backend.
   sellers: Tuple<Seller>|False with: (as-method => true)
   resolved-price: Double|False with: (as-method => true)
 with :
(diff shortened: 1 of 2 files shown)


bhusang (approved):

PhilippeFerreiraDeSousa:
To fix the UI issue
image

you need to modify this code in design-explorer/api.stanza
```
  defn to-design-explorer-ir (v: PropertyValue) -> TreePropertyValue :
    match(v) :
      (v: ClientValue) :
        val object =
          if contains?(["Pin", "Net"], name(v)) and (length(values(v)) == 1) :
            get?(object-by-local-id-table, id(values(v)[0] as Local))

        match(object: FlatPin|FlatNet) :
          object
        else :
          TreeClientValue(name(v), map(to-design-explorer-ir, values(v)))
      (v) :
        v
```
to handle tuples with:
```
      (v: Tuple<PropertyValue>) :
        map(to-design-explorer-ir, v)
```

Then extend
```
public deftype TreePropertyValue :
  PropertyValue <: TreePropertyValue
  TreeClientValue <: TreePropertyValue
  FlatPin <: TreePropertyValue
  FlatNet <: TreePropertyValue
```
to accept tuples of itself with:
```
  Tuple<T> <: TreePropertyValue
```

utils/parts.stanza
```diff
@@ -495,9 +495,21 @@ defn NoConnectCode (json: JObject) -> NoConnectCode :
...
+  to-jitx(c, false)
+
+; Set sellers as a property so they show up in the component card.
+public defn set-sellers-as-property (sellers: Tuple<Seller>|False) :
+  match(sellers:Tuple<Seller>) :
+    inside pcb-component :
+      var sellers-str = ""
+      for seller in sellers do :
+        sellers-str = append-all([sellers-str, "\n", to-string(seller)])
+      set-property(self, `sellers, sellers-str)
```
PhilippeFerreiraDeSousa, line 510: This should be
```
; Set sellers as a property so they show up in the component card.
 public defn set-sellers-as-property (sellers: Tuple<Seller>|False) :
  match(sellers:Tuple<Seller>) :
    property(self.sellers) = sellers
```

PhilippeFerreiraDeSousa: The stanza way to write your original code would have been `to-string("%n" % [sellers])` or `string-join(sellers, "\n")`

ptuckey: Thanks @PhilippeFerreiraDeSousa!

Have updated with this (and above changes in jitx-client), and re-tested.

bhusang (approved):

PR #405, opened 2023-06-27
Improve bypass-cap-strap

hypadr1v3:


utils/generic-components.stanza
```diff
@@ -585,12 +585,18 @@ public defn gen-led-cmp (pkg-name:String) :
 ; ========================================================
 ; These functions attach a passive component between two pins. e.g. strap a pull up resistor between an IO and VDD
 
-public defn bypass-cap-strap (power-pin:JITXObject, gnd-pin:JITXObject, capacitance:Double|Toleranced):
+public defn bypass-cap-strap (power-pin:JITXObject, gnd-pin:JITXObject, params:Tuple<KeyValue>) -> JITXObject :
```
falstjit, line 588: stanza now supports keyword arguments, could I suggest that we use those instead here?

hypadr1v3: Sure, I'll look into that.

hypadr1v3: Is keyword arguments documented anywhere? I can't seem to find anything which would help me know the syntax of this

PR #404, opened 2023-06-27
JITX-5312/code generator

erwin-lau:
### Changes
* add a function to create component file and change stanza.proj
* add a function to convert parts-db ComponentCode into JITX Importer FileCode for code-generator to create stanza code

### This PR is to merge to `dev`. Need help to make sure it is pinned to jitx-client and released to `main`.

PR #406, opened 2023-06-29
JITX-5377 / JITX-5376: Restore queries after parts-db fix deployed

ptuckey:
## Summary
We partially disabled some test queries, in order to unblock a PR from merging:
- #404

The root cause was a new version of the parts database, with some bad data:
- stock (zero values incorrectly set to -1)
- metadata.applications (missing)

Those issues have been fixed, so we're now restoring the original queries.

## Test Plan
I expect the OCDB CI to fail initially, as the fixes aren't fully deployed.

The population is running overnight, and I'll re-trigger the CI in the morning, and it should pass.

### DIFF
--- a/designs/doc-examples.stanza
+++ b/designs/doc-examples.stanza
@@ -121,7 +121,7 @@ pcb-module example-instances :
   inst c12 : ceramic-cap(["capacitance" => closest-std-val(test-voltage * 1.0e-6, 20.0) "min-rated-voltage" => 60.0])
 
 ; c13 = the largest value capacitor in case size 0805, 100V minimum rated voltage, and `metadata.applications` matches one of the fields in the `automotive-metadata` value.
-  inst c13 : ceramic-cap(["_sort" => ["-capacitance"] "case" => "0805" "min-rated-voltage" => 30.0 "metadata.applications" => automative-metadata])
+  inst c13 : ceramic-cap(["capacitance" => 68.0e-12 "_sort" => ["-capacitance"] "case" => "0805" "min-rated-voltage" => 30.0 "metadata.applications" => automative-metadata])
 ; ANCHOR_END: capacitor-examples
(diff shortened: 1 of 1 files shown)


designs/doc-examples.stanza
```diff
@@ -114,18 +114,14 @@ pcb-module example-instances :
...
 
 ; c12 =  ceramic capacitor who's value is calculated based on a capacitance * a double, then rounded to the nearest 20% standard value, 60V minimum rated voltage.
   val test-voltage = 5.0
   inst c12 : ceramic-cap(["capacitance" => closest-std-val(test-voltage * 1.0e-6, 20.0) "min-rated-voltage" => 60.0])
 
 ; c13 = the largest value capacitor in case size 0805, 100V minimum rated voltage, and `metadata.applications` matches one of the fields in the `automotive-metadata` value.
-  ; TODO: JITX-5376 - Restore "metadata.applications" filter.
-  inst c13 : ceramic-cap(["_sort" => ["-capacitance"] "case" => "0805" "min-rated-voltage" => 30.0])
-  ; inst c13 : ceramic-cap(["_sort" => ["-capacitance"] "case" => "0805" "min-rated-voltage" => 30.0 "metadata.applications" => automative-metadata])
+  inst c13 : ceramic-cap(["_sort" => ["-capacitance"] "case" => "0805" "min-rated-voltage" => 30.0 "metadata.applications" => automative-metadata])
```
ptuckey, line 124: This call is failing CI still:
```
FATAL ERROR: Key (capacitance) not found in
```

It's because the dummy substitution fallback (find-subs-comp) expects the "capacitance" field to be set, which isn't the case here.

I'll update to set a "capacitance" for this.

Separately, it seems that it previously wasn't an issue, because we were returning a value - it may be that the "metadata.applications" isn't getting set properly anymore, despite an intended fix in the population logic. I'll keep that ticket open to investigate that aspect more.

erwin-lau (approved):
looks right

PR #408, opened 2023-06-29
refactor code

erwin-lau:
### Background

Since code generation from REPL may not be ready any time soon, it is not necessary to expose importer2 and code-generator in OCDB

### Changes
* create a new file in tools/explorer-code-generator
* move parts-db code generator code there


ptuckey (approved):
LGTM!

PR #409, opened 2023-07-05
Fix CI to work with parts-v1

ptuckey:
The updated `stock` numbers in `parts-v1` broke some of the existing test queries.

The previously-returned parts still exist, but with zero stock now (which can be expected),
so the query failed to find results.

The fix here is to relax the constraints on the OCDB test queries, so that parts with >= 500 stock are found.

## Test Plan
The CI on this PR was failing, but is now passing.


erwin-lau (approved):
Looks reasonable

PR #410, opened 2023-07-07
Remove jstanza - use generic stanza instead

jwatson0:


falstjit:
Does this change actually do what the title says it does? :thinking: It seems completely unrelated? :D I assume it's some merge that shows up and the changes are outside the PR?

jwatson0:
wrong branch

PR #414, opened 2023-07-08
JITX-5435: Update resonator-checks to work with Toleranced

ptuckey:
## Summary
The parts-db population has been inserting "tolerance" into ESIR properties as a "Toleranced" defstruct.

It was previously represented as a scalar, using the max value, so we changed it to prevent losing data,
but this broke existing usage in OCDB's `resonator-checks`.

The fix here updates `resonator-checks` to work with either a `Toleranced` or a `Double`.

### 1. Run the `comprehensive-checks.stanza` design:
- This check actually manually set `tolerance` to a 2-tuple, which may have been another older format.
- Rather than support all 3 formats, this PR updates that check to use a Toleranced instead.

### 2. Test via OCDB's nRF52840 module
Add this to a `main.stanza`, and run:
```
set-main-module(ocdb/components/nordic/nRF52840/module)
run-check-on-design(ocdb/components/nordic/nRF52840/module)
```
- This contains `check-resonator` calls, which use a capacitor from the parts-db (via `add-xtal-caps` -> `cap-strap`).
- This failed without the PR changes, but passes with them.

### DIFF
--- a/designs/comprehensive-checks.stanza
+++ b/designs/comprehensive-checks.stanza
@@ -187,7 +187,7 @@ pcb-component crystal-resonator :
 pcb-module resonator-module :
   public inst X1 : crystal-resonator
   public inst C1 : ceramic-capacitor
-  property(C1.tolerance) = [0.99, 1.01]
+  property(C1.tolerance) = Toleranced(0.99, 1.0, 1.01)
   val intf = CrystalOscillator(
     max-critical-gain, drive-level,
     c-pin, frequency-tolerance, frequency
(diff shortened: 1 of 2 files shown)


designs/comprehensive-checks.stanza
```diff
@@ -187,7 +187,7 @@ pcb-component crystal-resonator :
 pcb-module resonator-module :
   public inst X1 : crystal-resonator
   public inst C1 : ceramic-capacitor
-  property(C1.tolerance) = [0.99, 1.01]
+  property(C1.tolerance) = Toleranced(0.99, 1.0, 1.01)
```
ptuckey, line 190: The actual values within Toleranced match a proposal discussed in JITX-5427.

I've gone with the `min/typ/max` proposed there - but, can update here to match whatever is decided.

erwin-lau (approved):
Go!

bhusang (approved):

PR #415, opened 2023-07-14
JITX-5504: Remove "sellers" ESIR property

ptuckey:
This pairs with a parts-db PR:
- [JITX-5504: Remove non-useful ESIR properties #142](https://github.com/JITx-Inc/parts-db/pull/142)

The "sellers" property was being inserted here in OCDB, since it may have been updated by `jit-backend`.

This is no longer required, as `sellers` will only be on the defstructs.

## Test Plan
See parts-db PR.

### DIFF
--- a/utils/db-parts.stanza
+++ b/utils/db-parts.stanza
@@ -150,7 +150,7 @@ defmethod to-jitx (resistor: Resistor) -> Instantiable :
   match(landpattern(component)) :
     (lp:LandPatternCode):
       ; If the landpattern is included, this part should have everything it needs.
-      to-jitx(component, sellers(resistor))
+      to-jitx(component)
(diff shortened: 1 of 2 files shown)


callendorph (approved):
LTGM
As with the related PR - we should probably hold off merging until we figure how we will present ephemeral data in the UI.

callendorph:
A lot of the test failures seem to be related to server backend issues. @ptuckey  is that typical ?

ptuckey:
> A lot of the test failures seem to be related to server backend issues. @ptuckey is that typical ?

Not expected! Looking into it...

bhusang:
@erwin-lau Do you want to do anything with this PR?

PR #420, opened 2023-07-21
JITX-5596/support-min-max

erwin-lau:
### Change

* support min-max code from JSON to create a Toleranced value.
* with MinMax in JSON, this paves the way to remove typical value in case in the future.
* more PR in jitx-client and parts-db to come.

### DIFF
--- a/utils/parts.stanza
+++ b/utils/parts.stanza
@@ -181,6 +181,12 @@ public-when(TESTING) defstruct TolerancedCode :
 with :
   printer => true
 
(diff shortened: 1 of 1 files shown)


bhusang (commented):
Should we also have a `TolerancedMinTypMaxCode()`? I think so.

erwin-lau:
> Should we also have a `TolerancedMinTypMaxCode()`? I think so.

The original `TolerancedCode` is indeed TolerancedMinTypMaxCode` with a computation.

erwin-lau:
Revert OCDB change

PR #426, opened 2023-08-06
simply changed all parametric-* references to placeholder-*

bhusang:


callendorph (approved):
:+1: LGTM
Do we need to check any of the cookbook or other projects for explicit `parametric-*` usage ?

bhusang:
I will do the check for usage as well.

PR #428, opened 2023-08-08
changed dummy-* dev to accept package size

bhusang:


utils/placeholder-components.stanza
```diff
@@ -41,52 +41,59 @@ public pcb-enum /ocdb/utils/placeholder-components/CapacitorApplication :
 ; ; ========== Generic passive components ==================
 ; ; ========================================================
 
-public pcb-component dummy-resistor (res:Double|Toleranced orig:Double|Toleranced|False = false):
+defn select-case (case:String|False) -> String :
+  var chosen-case = MIN-PKG
+  if case is-not False :
+    if is-valid-pkg(case as String) :
+      chosen-case = case
+  chosen-case as String
```
callendorph, line 49: This feels like it might be a common enough case to include in `generic-components.stanza`

A slightly more idiomatic approach that avoids the mutable variable:

```
stanza> val b = ["0402", "0603", "0805"]
stanza> val MIN-PKG = "0402"
stanza> defn select-case (case:String|False) -> String : 
      >   match(case):
      >     (x:String): reduce({ if _1 == x : _1 else : _0}, MIN-PKG, b)
      >     (y:False): MIN-PKG
```

This will use `MIN-PKG` if no valid package is found. If a valid package is found - it will return that package string. If `false` - it will return `MIN-PKG`. 

The `{ if _1 == x : _1 else : _0}` seems like a really common operation that would need to be done but I'm not seeing anything in the reference about a "if-match?-or-default" style function. 

I don't have a strong opinion here - so don't feel like you have to make this change.

callendorph (approved):
👍 

I made one minor suggestion - feel free to disregard.

utils/placeholder-components.stanza
bhusang: great suggestion. changed and committed.

PR #429, opened 2023-08-09
Purge Extraneous TRACESWO definition from SWD definition

callendorph:
This updates the SWD bundle to remove an extraneous signal. 

I've checked the cookbook and there are no instances of this there. 

@bhusang  - Is there anywhere else I should check for code that might touch the SWD bundle ?

### DIFF
--- a/components/tag-connect/TC2030-IDC-NL.stanza
+++ b/components/tag-connect/TC2030-IDC-NL.stanza
@@ -49,3 +49,10 @@ public pcb-component component :
     swd-swo().swdio  => self.p[2]
     swd-swo().swdclk => self.p[4]
     swd-swo().swo => self.p[6]
+
+  supports power:
+    power.vdd => self.p[1]
+    power.gnd => self.p[5]
+
+  supports reset:
+    reset.reset => self.p[3]
--- a/components/tag-connect/TC2030-IDC-NL.stanza
+++ b/components/tag-connect/TC2030-IDC-NL.stanza
@@ -45,8 +45,7 @@ public pcb-component component :
   landpattern = TC2030-IDC-NL-footprint(for i in 1 through 6 do : p[i] => TC2030-IDC-NL-footprint.p[i])
   reference-prefix = "J"
 
-  supports swd([SWD-SWO, SWD-TRACESWO]) :
-    swd([SWD-SWO, SWD-TRACESWO]).swdio  => self.p[2]
-    swd([SWD-SWO, SWD-TRACESWO]).swdclk => self.p[4]
-    swd([SWD-SWO, SWD-TRACESWO]).swo => self.p[6]
-    swd([SWD-SWO, SWD-TRACESWO]).traceswo => self.p[1]
+  supports swd-swo() :
+    swd-swo().swdio  => self.p[2]
+    swd-swo().swdclk => self.p[4]
+    swd-swo().swo => self.p[6]
--- a/components/tag-connect/TC2030-IDC.stanza
+++ b/components/tag-connect/TC2030-IDC.stanza
@@ -54,11 +54,10 @@ public pcb-component component :
 
   property(self.rated-temperature) = false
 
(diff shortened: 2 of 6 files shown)


bhusang (approved):
Looks good. We should also check the STM32 code for the swd() usage but I think that they are all good with this change.

bhusang:
It turns out that the STM32 code is throwing an error in CI:
```
jitx-client_1  | Running can-stm32.stanza...
jitx-client_1  | FATAL ERROR: No appropriate switch clause.
jitx-client_1  |   in ocdb/utils/bundles/get-bundle-by-name
jitx-client_1  |     at /app/open-components-database/utils/bundles.stanza:357.10
jitx-client_1  |   in ocdb/utils/bundles/get-bundle-by-name
jitx-client_1  |     at /app/open-components-database/utils/bundles.stanza:355.28
jitx-client_1  |   in ocdb/utils/stm-to-jitx/to-jitx-supports
jitx-client_1  |     at /app/open-components-database/utils/stm-to-jitx.stanza:157.10
jitx-client_1  |   in ocdb/utils/stm-to-jitx/to-jitx-supports
jitx-client_1  |     at /app/open-components-database/utils/stm-to-jitx.stanza:148.26
jitx-client_1  |   in ocdb/utils/micro-controllers
jitx-client_1  |     at /app/open-components-database/utils/micro-controllers.stanza:524.2
jitx-client_1  |   in ocdb/utils/stm-to-jitx/mcu-module
jitx-client_1  |     at /app/open-components-database/utils/stm-to-jitx.stanza:170.16
jitx-client_1  |   in ocdb/designs/can-stm32
jitx-client_1  |     at can-stm32.stanza:37.7
jitx-client_1  |   in ocdb/designs/can-stm32
jitx-client_1  |     at can-stm32.stanza:72.38
jitx-client_1  | An error occurred during execution.
```

callendorph:
Thanks - I'll check it out.

callendorph:
OK - I was able to replicate this issue locally. Here is the part info that I think is causing the problem. This is in MPN 

```json
                 
                            }, {
                                        "bundle":       {
                                                "name": "swd",
                                                "options":      ["SWD-TRACESWO"]
                                        },
                                        "mappings":     [{
                                                        "bundle-pin":   "swdio",
                                                        "require":      "SYS_JTMS-SWDIO",
                                                        "pin":  "p"
                                                }, {
                                                        "bundle-pin":   "swdclk",
                                                        "require":      "SYS_JTCK-SWCLK",
                                                        "pin":  "p"
                                                }, {
                                                        "bundle-pin":   "traceswo",
                                                        "require":      "SYS_JTDO-TRACESWO",
                                                        "pin":  "p"
                                                }]
                                }, {

```

It seems that the STM32 parts in the database may be using the `SWD-TRACESWO` instead of the `SWD-SWO` option. 

I think I can approach this in two ways: 

1.  Modify the database and convert the `SWD-TRACESWO` to `SWD-SWO`. 
2.  Modify the `ocdb` code in `get-bundle-by-name` to handle this legacy option. 

I'm going to implement option 2.

bhusang:
Are we sure that editing the database is not the best way to solve this?

callendorph:
I'm beginning to think that may be the better course of action.

I was concerned that changing the database would break ocdb for older versions, but now I'm convinced that is not true. I can change the `SWD-TRACESWO` to `SWD-SWO` and I can change the `traceswo` bundle pin to `swo` and this should be backward compatible with existing OCDB implementations because they all support the `SWD-SWO` variant.

callendorph:
OK - I've pushed a change into the backend that should fix this for the STM32 microcontrollers. Fortunately, I did not have to touch the database - I just had to modify some python code. 

I'll test that this fix causes the tests to succeed locally and then we can re-run the CI.

callendorph:
OK - I've force pushed an update that doesn't remove the TRACESWO yet so that we can get these changes merged while we wait to modify the database.

PR #430, opened 2023-08-09
Consistent interfaces for `cap-strap` and `bypass-cap-strap`

callendorph:
This adds a new interface to `bypass-cap-strap` to allow the user to pass `KeyValuePair` parameters similar to the base method for `cap-strap`.

### DIFF
--- a/utils/generic-components.stanza
+++ b/utils/generic-components.stanza
@@ -585,11 +585,17 @@ public defn gen-led-cmp (pkg-name:String) :
 ; ========================================================
 ; These functions attach a passive component between two pins. e.g. strap a pull up resistor between an IO and VDD
 
-public defn bypass-cap-strap (power-pin:JITXObject, gnd-pin:JITXObject, capacitance:Double|Toleranced):
+public defn bypass-cap-strap (power-pin:JITXObject, gnd-pin:JITXObject, params:Tuple<KeyValue>):
   inside pcb-module:
-    val cap = cap-strap(gnd-pin, power-pin, capacitance)
+    val cap = cap-strap(gnd-pin, power-pin, params)
     short-trace(cap.p[2], power-pin)
     cap
+
+public defn bypass-cap-strap (power-pin:JITXObject, gnd-pin:JITXObject, capacitance:Double|Toleranced, tol:Double):
+  bypass-cap-strap(power-pin, gnd-pin, ["capacitance" => capacitance, "tolerance" => tol])
+
+public defn bypass-cap-strap (power-pin:JITXObject, gnd-pin:JITXObject, capacitance:Double|Toleranced):
+  bypass-cap-strap(power-pin, gnd-pin, ["capacitance" => capacitance])
 
 public defn bypass-caps (power-pin:JITXObject|Symbol, gnd-pin:JITXObject|Symbol, voltage:Double, caps:Tuple<Double|Toleranced>, name:Symbol) :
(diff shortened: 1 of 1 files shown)


utils/generic-components.stanza
```diff
@@ -585,12 +585,18 @@ public defn gen-led-cmp (pkg-name:String) :
...
+    val cap = cap-strap(gnd-pin, power-pin, params)
     short-trace(cap.p[2], power-pin)
     cap
 
+public defn bypass-cap-strap (power-pin:JITXObject, gnd-pin:JITXObject, capacitance:Double|Toleranced, tol:Double):
+  bypass-cap-strap(power-pin, gnd-pin, ["capacitance" => capacitance, "tolerance" => tol])
+
+public defn bypass-cap-strap (power-pin:JITXObject, gnd-pin:JITXObject, capacitance:Double|Toleranced):
+  bypass-cap-strap(power-pin, gnd-pin, ["capacitance" => capacitance])
+
```
bhusang, line 599: So this might be a good place to use the undocumented optional/named arguments in Stanza.

The function could look something like (no guarantees that this code works 😸 ):
```
public defn bypass-cap-strap (power-pin:JITXObject, gnd-pin:JITXObject, -- params:Tuple<KeyValue> = false capacitance:Double|Toleranced|False = false, tol:Double|False = false)
  var my-params:Tuple<KeyValue>
  if params is-not False:
    add(my-params, params)
  if capacitance is-not False :
    add(my-params, ["capacitance" => capacitance])
  if tol is-not False:
    add(my-params, ["tolerance" => tol])
    inside pcb-module:
  val cap = cap-strap(gnd-pin, power-pin, params)
  short-trace(cap.p[2], power-pin)
  cap
```

bhusang: You would then need to call the function as
`bypass-cap-strap(pina, pinb, params=params-to-send)`
`bypass-cap-strap(pina, pinb, capacitance=1.0e-6)` 
and so on...

callendorph: Maybe as a way to not break the existing code - we could consider:

```
public defn bypass-cap-strap (power-pin:JITXObject, gnd-pin:JITXObject, capacitance:Double|Toleranced, tol:Double -- params:Tuple<KeyValue> = false):
  .. Handle the `params` arg here. 
  bypass-cap-strap(power-pin, gnd-pin, ["capacitance" => capacitance, "tolerance" => tol])

public defn bypass-cap-strap (power-pin:JITXObject, gnd-pin:JITXObject, capacitance:Double|Toleranced -- params:Tuple<KeyValue> = false):
  .. Handle the `params` arg here
  bypass-cap-strap(power-pin, gnd-pin, ["capacitance" => capacitance])
```

That would add the ability to include the parameters while keeping the positional argument versions intact. 

I think we need a plan for how we are going to update this interface. Just doing it adhoc PR to PR doesn't seem like a good idea. Better to rip the bandaid off and a have a plan for supporting legacy implementations.

bhusang: I agree with the concern to not upend existing usage. But unfortunately the only way that we are going to use the full power of the named and optional arguments is to rip the band-aid off.

callendorph:
Given that I don't think we are going to approach these larger refactors until we can create a package for OCDB, I think we should consider merging this for consistent interface in OCDB for now.

bhusang (approved):

PR #431, opened 2023-08-10
JITX-5813: Add new 3DModel fields from parts-db

ptuckey:
Parse new fields for 3D models from parts-db.

We weren't populating any 3D models before, so it's safe to assume the new fields exist.

## Test Plan
See [parts-db PR](https://github.com/JITx-Inc/parts-db/pull/181).

### DIFF
--- a/utils/parts.stanza
+++ b/utils/parts.stanza
@@ -972,8 +972,11 @@ defn Model3D (json: JObject) -> Model3D :
           Vec3D(json["position"] as JObject),
           Vec3D(json["scale"] as JObject),
           Vec3D(json["rotation"] as JObject),
-          get(json, "properties") as Tuple<KeyValue<String, String>>,
-          json["jitx_model_3d_id"] as String)
+          ; Serialization not yet supported for:
+          ;   Tuple<KeyValue<String, String>>
+          [],
+          json["jitx_model_3d_id"] as String,
+          json["comment"] as String)
 
 defn Vec3D (json: JObject) -> Vec3D :
   Vec3D(json["x"] as Double,
--- a/utils/parts.stanza
+++ b/utils/parts.stanza
@@ -1008,7 +1008,13 @@ defn to-jitx (lp: LandPatternCode, jitx-pads-by-pcb-pad-name: HashTable<String, 
 
     do(to-jitx, layers(lp))
     do(to-jitx, geometries(lp))
-    do(to-jitx, model3ds(lp))
+    ; TODO: JITX-5895 - Add 3D models for parametric queries
+    ;
+    ;       For now, we don't add 3D models, because we don't know the relative transformations.
+    ;
+    ;       If users want 3D models, they can use "Create component" in component search,
+    ;       which generates Stanza files, that include a comment about the relative transformations.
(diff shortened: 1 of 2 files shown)


bhusang (approved):
I assume there are corresponding changes in the Model3D api to support the new syntax?

ptuckey:
> I assume there are corresponding changes in the Model3D api to support the new syntax?

Yep - forthcoming jitx-client PR!

erwin-lau (approved):
LGTM!

utils/parts.stanza
```diff
@@ -971,7 +971,9 @@ defn Model3D (json: JObject) -> Model3D :
   Model3D(json["filename"] as String,
           Vec3D(json["position"] as JObject),
           Vec3D(json["scale"] as JObject),
-          Vec3D(json["rotation"] as JObject))
+          Vec3D(json["rotation"] as JObject),
+          get(json, "properties") as Tuple<KeyValue<String, String>>,
```
erwin-lau, line 975: I don't know this is possible.

ptuckey: I wasn't sure either, and I started using the approach from `VendorPartNumbers` - but, in testing, this one seemed to work.

bhusang:
You will need to have a jitx-client docker image with the Model3D() updated and then change the docker image name in this repo in order to get CI to pass. @jwatson0 can help to create the docker image, I believe.

utils/parts.stanza
```diff
@@ -1003,7 +1008,13 @@ defn to-jitx (lp: LandPatternCode, jitx-pads-by-pcb-pad-name: HashTable<String,
...
     do(to-jitx, layers(lp))
     do(to-jitx, geometries(lp))
-    do(to-jitx, model3ds(lp))
+    ; TODO: JITX-5895 - Add 3D models for parametric queries
+    ;
+    ;       For now, we don't add 3D models, because we don't know the relative transformations.
+    ;
+    ;       If users want 3D models, they can use "Create component" in component search,
+    ;       which generates Stanza files, that include a comment about the relative transformations.
+    ;do(to-jitx, model3ds(lp))
```
ptuckey, line 1017: @erwin-lau @bhusang Just wanted to highlight this change, which I added since you accepted.

More details on JITX-5895.

bhusang: We should highlight the components that contain 3D models and use that highlighting to hint to the user to use the `Create Component` option to get the 3D data where the user still needs to figure out the appropriate coordinate transform.

utils/parts.stanza
ptuckey: Update - although it worked within OCDB, it failed later on during serialization.

Have updated to remove the `properties`, and pass the new comment we need as a `comment` field.

erwin-lau:
Now this time it is OCDB depends on a definition from jitx-client

bhusang:
Yes, you have to change the jitx-client docker image in the .github sub-dir.

utils/parts.stanza
erwin-lau: I am gong to close the PR since the ESIR change will be not necessary

PR #433, opened 2023-08-17
removed usage of DualOffsetVoltage()

bhusang:


callendorph:
This seems like it removes the usages but doesn't remove the struct definition in [utils/property-structs.stanza](https://github.com/JITx-Inc/open-components-database/blob/51c6acf8451356da49801a9f054dbbfeb1f42763/utils/property-structs.stanza#L196)

Is that intentional?

bhusang:
Yes, for now I don't want to remove the struct because that should be part of a greater cleanup like removing RelativeVoltage().

PR #437, opened 2023-08-23
added board info to run test correctly

bhusang:


erwin-lau (approved):
Thank you. It was fast!

PR #447, opened 2023-09-20
added dnp to text & logo plus small design test

bhusang:
To remove the logo and text from the BOM, set dnp() on those components.

### DIFF
--- a/artwork/board-text.stanza
+++ b/artwork/board-text.stanza
@@ -20,6 +20,7 @@ defpackage ocdb/artwork/board-text :
   import ocdb/utils/defaults
   import ocdb/utils/landpatterns
   import ocdb/utils/generic-components
+  import ocdb/utils/generator-utils
   import ocdb/utils/symbols
 
 pcb-landpattern text-lp (text:String, size:Double, outline-size:Double) :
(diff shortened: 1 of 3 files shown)


erwin-lau (approved):
Thanks for your help.

artwork/board-text.stanza
```diff
@@ -52,7 +53,10 @@ public pcb-component text (title:String, size:Double, outline-size:Double) :
   symbol = sym()
   reference-prefix = "TXT"
   property(self.rated-temperature) = false
-  
+
 public defn version-silkscreen (title:String) :
-  text(title, 1.5, 0.1)
+  inside pcb-module :
+    inst my-text : text(title, 1.5, 0.1)
+    dnp(my-text)
```
callendorph, line 60: Are you sure these should be `dnp` and not `component-status(NotInBOM)` or something similar.

I think there is a subtle distinction. With DNP it is a component that is not stuffed and the schematic should have a `X` over it.

For the `NotInBOM` - I associate this with mechanical components like a mounting hole or logo where it should not be in the BOM - but it should also not have an `X` in the schematic.

That all being said - I don't think that `NotInBOM` is handled this way as of today.

bhusang: Ideally they should be NotInBOM status, however, setting that as an option was never exposed, so the only function we have is to query the status:
```
doc: \<doc>
  Retrieve the component in-BOM and on-board status of a component instance
<doc>
public defn component-status? (obj:JITXObject) -> [ComponentBOMStatus ComponentBoardStatus] :
  component-status?(JITX-CLIENT, obj)
```

I think that I will open a ticket to add the component statuses from user codespace.

PR #454, opened 2023-10-13
JITX-5504/no-sellers-in-properties

erwin-lau:
### Changes

* remove `sellers` from being an ESIR property.
* As such, sellers will not show up as one of the component properties

image
image

* if a part is selected in the design, the design-explorer will look like this.
image

### NOTE

* Since explorer depends on OCDB, a build with explorer as the target is necessary to be effective

### DIFF
--- a/utils/db-parts.stanza
+++ b/utils/db-parts.stanza
@@ -153,7 +153,7 @@ defmethod to-jitx (resistor: Resistor) -> Instantiable :
   match(landpattern(component)) :
     (lp:LandPatternCode):
       ; If the landpattern is included, this part should have everything it needs.
-      to-jitx(resistor, component, sellers(resistor))
+      to-jitx(resistor, component)
     (c):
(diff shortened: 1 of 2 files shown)


bhusang (approved):
Assuming that we add a tab to the component explorer for the seller information, ok to remove it here. This was the same constraint that I expressed to Pat when he wanted to remove the sellers property. We need the sellers info to be user viewable.

erwin-lau:
> Assuming that we add a tab to the component explorer for the seller information, ok to remove it here. This was the same constraint that I expressed to Pat when he wanted to remove the sellers property. We need the sellers info to be user viewable.

The server still return the sellers data. It is not put into `property(self.sellers)` anymore.
In parts resolution and in BOM generation, there is no change, the sellers are still in.
The tree view (design-explorer) is generated during parts resolution, to move it to another tab is just a UI change.

PR #455, opened 2023-10-16
reroute to ocdb/utils/fonts

erwin-lau:
* kicad-string-length should not be called by OCDB into jitx-client
* it is moved to OCDB now.

### DIFF
--- a/utils/box-symbol.stanza
+++ b/utils/box-symbol.stanza
@@ -6,7 +6,7 @@ defpackage ocdb/utils/box-symbol :
   import lang-utils
   import jitx
   import jitx/commands
-  import jitx/fonts
+  import ocdb/utils/fonts
 
 ;============================================================
 ;========== Definition of a Parameterized Box Symbol ========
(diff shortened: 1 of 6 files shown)


PhilippeFerreiraDeSousa:
Yes, however everybody's design won't run anymore on "can't find package jitx/fonts" in a week when they get auto-updated.

I thought we could wait for a meaningful breaking change to include this in. I just forgot jitx/fonts would not be built anymore in release.

erwin-lau:
> Yes, however everybody's design won't run anymore on "can't find package jitx/fonts" in a week when they get auto-updated.
> 
> I thought we could wait for a meaningful breaking change to include this in. I just forgot jitx/fonts would not be built anymore in release.
This includes an `ocdb/utils/fonts` so that new project will work.
For old designs to work, we should restore the package in `jitx-client`.

PhilippeFerreiraDeSousa (approved):
Good to do both, do that for new project created and Jason put back jitx/fonts in jitx for old projects

PR #458, opened 2023-10-27
JITX-6927/shorten approved distributor list

erwin-lau:
### Changes

* remove 4 vendors from APPROVED-DISTRIBUTOR-LIST
* DigiKey is ordered first, followed by Mouser


bhusang:
If we remove these disti's, especially JLCPCB, does that affect the parts lookup materially?

erwin-lau:
> If we remove these disti's, especially JLCPCB, does that affect the parts lookup materially?

We don't remove them totally. Just remove them from the default list. User can still provide JLCPCB in the list they provide in `setup-project`.
From my anecdote, JLCPCB is harder to search, esp. when we put them first in the list because they have higher priority now. Cofactr, which we rely on for dynamic stock and price update, data does not carry JLCPCB information.

bhusang (approved):

PR #462, opened 2023-11-16
Add default vias

d-haldane:


utils/defaults.stanza
```diff
@@ -65,17 +65,37 @@ public pcb-rules default-rules :
   min-pth-pin-solder-clearance = 3.0
 set-default-rules(default-rules)
 
+public pcb-via default-via :
+  name = "Default TH"     
+  start = LayerIndex(0, Top)
+  stop = LayerIndex(0, Bottom)
+  diameter = 0.6
+  hole-diameter = 0.3
+  type = MechanicalDrill
```
bhusang, line 74: May want to add tented = true by default (I think that most vias are tented by default?)

bhusang:
So this PR is bound to not pass the CI due to needing a different version of jitx-client than jitx-client:2.24.0.
@jwatson0 should create a new jitx-client docker image for 2.27 (when that is ready). Then that docker image will be referenced in the ocdb Dockerfile and then CI will pass.

utils/defaults.stanza
bhusang: https://jlcpcb.com/help/article/122-Via-Covering:-Tented,-Untented,-Plugged,-Epoxy-Filled-and-Copper-epoxy-filled 

tldr:
Tented Vias
This option is the most commonly used. Vias are covered with soldermask, and surface finish such as HASL is not applied to vias.

d-haldane: If that isn't the default value for a pcb-via then we should make it the default value. 

This PR is canonical, reality needs to be updated.

callendorph (approved):
I agree with tenting being a good default.

0.3mm for the default drill size is good because I think JLCPCB's min drill size without extra charges is 0.3mm. [Advanced Circuits's standard spec](https://www.4pcb.com/pcb-manufacturing-custom-standard.html) is 10mil => 0.25mm

I'm assuming that this has to wait until the behavior that supports the `vias =` syntax is present in JITX core.?

d-haldane:
> I'm assuming that this has to wait until the behavior that supports the `vias =` syntax is present in JITX core.?

That's right. Vias has been merged to develop so the release is soon. We'll need this in for that release.

bhusang:
@ariel-hi is adding the defaults for the vias into jitx-client. When we have a docker image ready, then this PR should go through.

PR #467, opened 2023-12-19
Add SLM Config

callendorph:
This adds an `slm.toml` configuration for OCDB. Right now, this doesn't do anything. It's primary function is support the integration of SLM in the JITX deployment. 

@jwatson0  - The `slm.toml` file has a version number in it. Ideally, that version number is updated with every release. I realize that may add some complexity to generating OCDB releases - but given the way that we are deploying, the file needs to be in the repo. 

Alternatively, we just leave the version and increment it when we add/remove functionality.

I'm open to other suggestions or ways to do this if you think there is a better way.


jwatson0:
- Can this be a full semver, like `2.29.0-develop.1`?
- Currently, we often have multiple jitx versions using the same ocdb commit, such as  `2.28.0-develop.1` and `2.28.0`. Which version should be in the `slm.toml`?
- Maybe the `slm.toml` should only be updated in the `main` branch for jitx releases?

jwatson0:
In order for the `jitx-client` submodule to be pointing to the correct `ocdb` commit that contains a `slm.toml` file with the correct version, we would need to do this in order:
- update ocdb slm.toml with the correct version, check in, probably bypass PR since there would be no code change.
- This generates a new commit in ocdb
- Update the submodule in jitx-client to point to the new ocdb commit
- Build jitx-client release

I don't see how we can do that for full semver versions inside `slm.toml` because the full jitx semver is not known until CI build time.

callendorph:
> Can this be a full semver, like 2.29.0-develop.1?

Unfortunately no. The [current implementation](https://github.com/StanzaOrg/semver) is very rigid and doesn't support this. Certainly something we can work on in the future.  

> Maybe the slm.toml should only be updated in the main branch for jitx releases?

Ya - I think that is probably what we should do if we are going to modify the file in OCDB.

> I don't see how we can do that for full semver versions inside slm.toml because the full jitx semver is not known until CI build time.

OK - so keeping them in lock step version might be hard. It feels like there are two options:

1.  We let the OCDB and jitx versions diverge - this requires some additional work to allow each release of JITX to know what version of OCDB to use. Not impossible, but will require some code changes.
2. We change nothing and we just insert a `slm.toml` file into the OCDB clone on the front end. It has its version set to whatever the release version is. The git hash/tag is the main driver for the version sync between OCDB/JITX and the slm.toml file is just there to make the slm dependency resolution work.

jwatson0:
Maybe instead of the `slm.toml` claiming that "this is ocdb version 2.29.0", it should claim "this ocdb requires jitx-client version 2.29.0".  Then, ocdb can lag slightly behind jitx in versioning, and the ocdb `slm.toml` version can be increased when a feature requiring a newer jitx version is introduced to ocdb.

If we keep the idea of slm.toml containing the ocdb version and require the ocdb version to exactly match the jitx-client version 1:1, then I don't think we could keep ocdb as a submodule of jitx-client without a large refactor of the build versioning system.  The CI build versioning system would need to create new commits in both repos during the build process, which it currently doesn't do.

jwatson0:
> let the OCDB and jitx versions diverge
So if we have jitx version 2.29.0 and ocdb version 1.5, do we say that "jitx requires ocdb 1.5" or "ocdb requires jitx 2.29.0"?  I think it has to be the latter.  So, the jitx client doesn't have pre-knowledge of which ocdb version to use, but instead finds the latest (?) ocdb version that is compatible with its current client version.

> just insert a slm.toml file into the OCDB clone on the front end.
I think that's a much easier solution.

- another option is to have a tiny `slm-ocdb` repo that only contains the `slm.toml` with the appropriate version and an ocdb submodule pointing to the appropriate commit.

callendorph:
> > just insert a slm.toml file into the OCDB clone on the front end.
> I think that's a much easier solution.

Ya - not great, but probably the easiest thing to do right now. I'll close this then and we will handle it in the jitx-client integration.

PR #470, opened 2024-01-08
scripts/run-ocdb-tests.sh: remove version-specific python invocation

jwatson0:


scripts/run-ocdb-tests.sh
```diff
@@ -9,7 +9,7 @@ if [ "$(basename $PWD)" != "open-components-database" ] ; then
     exit -1
 fi
 
-PYTHON=python3.10
+PYTHON=python
```
falstjit, line 12: may I suggest `PYTHON=${PYTHON:-python}` instead for easy override?

jwatson0: Sure, I did that in other scripts but missed this one.

bhusang (approved):

PR #479, opened 2024-03-28
Hash Locals instead of JITXObjects

ariel-hi:
A recent object core change ([linked here](https://github.com/JITx-Inc/jitx-client/pull/3983/files#r1543790951)) removes Hashable from JITXObject.
`ble-mote` crashes during runtime because it calls `add-testpoint`. I changed `hash` call to operate on the JITXObject's locals.

Also double checking @tjknoth did we want to remove Hashable?

### DIFF
--- a/utils/generic-components.stanza
+++ b/utils/generic-components.stanza
@@ -928,7 +928,7 @@ public pcb-module test-points (count:Int, method:ocdb/utils/generic-components/T
 
 public defn add-testpoint (rs:Tuple<JITXObject>, method:ocdb/utils/generic-components/Testpoint|Instantiable) :
   inside pcb-module :
-    val unique-id = symbol-join([`TP- to-int(hash(local(rs)))])
+    val unique-id = symbol-join([`TP- to-int(hash(map(local, rs)))])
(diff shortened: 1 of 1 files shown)


utils/generic-components.stanza
```diff
@@ -928,7 +928,7 @@ public pcb-module test-points (count:Int, method:ocdb/utils/generic-components/T
 
 public defn add-testpoint (rs:Tuple<JITXObject>, method:ocdb/utils/generic-components/Testpoint|Instantiable) :
   inside pcb-module :
-    val unique-id = symbol-join([`TP- to-int(hash(rs))])
+    val unique-id = symbol-join([`TP- to-int(hash(map(local, rs)))])
```
bhusang, line 931: @tjknoth Is there an easier way in the new object core to create a non-colliding id for the `make-inst()` function?

tjknoth: Can you just use `genid()`?
There is also `index(local)` which is defined on object-core/Object -- but it is probably not exposed to users. We can do so if need be.

bhusang (approved):

PR #480, opened 2024-03-29
PRD-52/Altium-GND-symbol

erwin-lau:
* use JITX symbol for Altium GND in (tick)altium backend.
* Altium export will still shown Altium GND.
image

### DIFF
--- a/utils/symbols.stanza
+++ b/utils/symbols.stanza
@@ -103,12 +103,11 @@ public pcb-symbol altium-power-gnd-power-sym :
   name = "ALTIUM-POWER-GND-POWER"
   pin p[0] at unit-point(0.0, 0.0)
 
-  unit-line([[0.0, 0.0], [1.0, 0.0]])
-  unit-line(0.1, [[1.0, -1.0], [1.0, 1.0]])
-  unit-line(0.1, [[1.28, -0.72], [1.28, 0.72]])
-  unit-line(0.1, [[1.56, -0.44], [1.56, 0.44]])
-  unit-line(0.1, [[1.84, -0.16], [1.84, 0.16]])
-  unit-val([1.0, -0.4])
+  unit-line([[0.0, 0.0], [0.5, 0.0]])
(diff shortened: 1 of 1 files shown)


ariel-hi:
This change has a few ramifications on the internal design flow.

If the Altium GND symbol is changed in JITX, this means we no longer need to keep any of the Altium-specific pcb-symbols in sync with how they appear in Altium.

We have a step in our workflow where symbols are substituted to Kicad or Altium form. I expect that substitution is no longer necessary.

d-haldane:
This is not the right fix for this problem. The issue was with `ground-sym` the JITX schematic visualizer ignores the graphical information in `ground-sym` and incorrectly shows the backend-subsitution symbol instead. 

It needs to show the geometry in ground-sym.

erwin-lau:
> We have a step in our workflow where symbols are substituted to Kicad or Altium form. I expect that substitution is no longer necessary.

If we eliminate substitution, at what point Altium knows the symbol needs to rotate 90 degrees clockwise to point downward?

erwin-lau:
> This is not the right fix for this problem. The issue was with `ground-sym` the JITX schematic visualizer ignores the graphical information in `ground-sym` and incorrectly shows the backend-substitution symbol instead.
> 
> It needs to show the geometry in ground-sym.

I believe the whole purpose of backend-substitution is to show what user should see in the selected CAD tool when it is exported. That is why the Altium ground symbol is a bit bigger because it is how it is drawn in Altium. This cannot be altered. Thus we mock Altium's drawing in our own form.

If we skip backend-substitution in JITX-visualizer we need to replace it with a step in exporter to rotate the ground symbol 90 degrees clockwise (and power symbols CCW) because all net-symbols in Altium point to the right.
There is another risk. Since the JITX-ground-symbol is smaller, it is possible to have overlapped drawings when the schematic is exported. (see the exported picture above, right). I am not sure if we have a mechanism to assign larger symbol bounds instead of calculates it to avoid potential overlapping.

erwin-lau:
Withdraw this until we have a better solution.

PR #482, opened 2024-04-08
use unsafe-hash

tjknoth:
uses the `unsafe-hash` function that I introduce in jitx-client PR 4068

### DIFF
--- a/utils/checks.stanza
+++ b/utils/checks.stanza
@@ -382,7 +382,7 @@ public defn propagate-power-states (module:JITXObject) :
   
   ; DFS walk of the netlist, to update power-states of 
   ; connected power-pins 
-  val visited = HashSet<JITXObject>()
+  val visited = HashSet<JITXObject>(unsafe-hash, equal?)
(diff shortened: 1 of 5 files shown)


bhusang:
This PR will need to wait until the jitx-client PR 4068 is completed and a new docker container is ready. Given that many of the cookbook recipes will fail without this PR which depends on the right docker container, we might need to create the container sooner than the release. @jwatson0 thoughts?

jwatson0:
Historically, we haven't created a new container instance with a pre-release jitx-client version because this repo is public and shouldn't rely on non-public content.  Also, this repo's build credentials do not have access to non-public pre-release jitx-client build files or containers.
We could either:
1) wait for jitx-client PR 4068 to be available in a public release (1-2 weeks)
2) commit this PR without a passing test build (skip the use of the docker container)
3) make a public container image of a pre-release version of jitx-client for this PR to run its tests

tjknoth:
The jitx-client PR has been merged to develop.

bhusang (approved):

PR #484, opened 2024-04-29
added SPI RDY

bhusang:
A new extension (maybe proprietary to STM?) is to add a RDY pin for the slave to indicate status. 

https://www.st.com/resource/en/application_note/an5543-guidelines-for-enhanced-spi-communication-on-stm32-mcus-and-mpus-stmicroelectronics.pdf

### DIFF
--- a/utils/bundles.stanza
+++ b/utils/bundles.stanza
@@ -153,17 +153,19 @@ public pcb-enum ocdb/utils/bundles/SPIPins :
   SPI-CIPO
   SPI-COPI
   SPI-SDIO
+  SPI-RDY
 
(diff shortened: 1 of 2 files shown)


utils/bundles.stanza
```diff
@@ -153,17 +153,19 @@ public pcb-enum ocdb/utils/bundles/SPIPins :
   SPI-CIPO
   SPI-COPI
   SPI-SDIO
+  SPI-RDY
```
callendorph, line 156: Seems very specific to the STM peripheral. I think it is fine to include it, but maybe we use a separate function to create SPI bundles with this pin include.

callendorph (approved):
:+1:

PR #494, opened 2024-08-01
use new keywords

erwin-lau:
* use new keywords `draw`, `stack` and `port`. 
* test: run `jstanza run-test ocdb/tests/all` and look for the word `deprecated`. Expected result: no deprecated WARNING


bhusang:
This requires a docker image that does not exist currently? It should target 3.14+ which means that this PR should only be merged after 3.14 is released. Or it can be merged with bypass branch protections.

erwin-lau:
> This requires a docker image that does not exist currently? It should target 3.14+ which means that this PR should only be merged after 3.14 is released. Or it can be merged with bypass branch protections.

I have done `ocdb/tests/all` locally and it ran ok. Nevertheless we can wait.
The consequence will be 3.14 will get lots of warning messages during run. They will go away in 3.15

erwin-lau:
@jwatson0 `docker compose` is used here instead of `docker-compose` to make this PR pass. Let me know if it is not appropriate.

AashnaP22 (approved):

PR #496, opened 2024-08-16
Tk/delta resistance

tjknoth:
Reopening my PR against correct base. Merged, so slm.toml is up-to-date.
I'll probably just force push this once it's approved so that my jitx-client PR goes through


plasmajitx (approved):
Yes, agreed with the force-pushing approach.

PR #504, opened 2025-02-14
Add dynamic via to via defs on default board

PhilippeFerreiraDeSousa:


d-haldane:
I don't think we should make this change - removing NFPs on vias is fairly advanced, sometimes unintuitive, and should not be int he default configuration.

